Issue #23, opened 2019-05-31
fix "similar-code" issue in projects/angular/components/src/ui-grid/managers/filter-manager.ts


vladjerca:
https://codeclimate.com/github/UiPath/angular-components/projects/angular/components/src/ui-grid/managers/filter-manager.ts?from_sha=d86bbb7e&to_sha=c3bf5d8f

Issue #20, opened 2019-05-31
Fix "similar-code" issue in projects/angular/directives/src/ui-dateformat/ui-dateformat.directive.ts


vladjerca:
Similar blocks of code found in 2 locations. Consider refactoring.

https://codeclimate.com/github/UiPath/angular-components/projects/angular/directives/src/ui-dateformat/ui-dateformat.directive.ts#issue_5cf1a6b4883fdb00010000c5

Issue #19, opened 2019-05-31
Fix "similar-code" issue in projects/angular/components/src/ui-grid/managers/filter-manager.ts


vladjerca:
Similar blocks of code found in 2 locations. Consider refactoring.

https://codeclimate.com/github/UiPath/angular-components/projects/angular/components/src/ui-grid/managers/filter-manager.ts#issue_5cf1a6b4883fdb00010000c1

Issue #18, opened 2019-05-31
Fix "similar-code" issue in projects/angular/directives/src/ui-secondformat/ui-secondformat.directive.ts


vladjerca:
Similar blocks of code found in 3 locations. Consider refactoring.

https://codeclimate.com/github/UiPath/angular-components/projects/angular/directives/src/ui-secondformat/ui-secondformat.directive.ts#issue_5cf1a6b4883fdb00010000b9

PR #30, opened 2019-06-03
Refactor to eliminate code smells and configure codeclimate

vladjerca:
# Fixed:

- #23 
- #22
- #20
- #19
- #18 

# Code Climate

- I've increased the line threshold to better suite our coding style, considering that we multi-line a lot to help with `merge conflicts`

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -4,3 +4,11 @@ exclude_patterns:
 - '**/*.spec.ts'
 - '**/*.d.ts'
 - '**/node_modules/'
+
+checks:
+  method-lines:
+    config:
+      threshold: 50
+
+  file-lines:
+    enabled: false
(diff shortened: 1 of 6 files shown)


.codeclimate.yml
```diff
@@ -4,3 +4,12 @@ exclude_patterns:
...
 - '**/node_modules/'
+
+checks:
+  method-lines:
+    config:
+      threshold: 50
+
+  file-lines:
+    config:
+      threshold: 500
```
sergiubologa: Isn't this too much? As I can see it's default value is 250.

vladjerca: @sergiubologa

Clearly we should increase it, due to the fact that we multiline a lot.

I'm just waiting for proposed values from you guys.

sergiubologa: I would suggest disabling this rule. As I can see the `ui-grid.component.ts` file has almost 600 lines and I suspect is not the largest file. Also, I can foresee even larger files in the future, because of the same reason you said: _we multiline a lot_.

What do you think?

vladjerca: That's a good suggestion. I've disabled the `file-line` count check.

PR #43, opened 2019-07-16
Feature / add snackbar component

sergiubologa:
- create ui-snackbar component
- create nl2br pipe
- add snackbar and pipes to commit scope list


projects/angular/components/ui-snackbar/src/ui-snackbar.component.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+                const snack5000 = getSnack();
+                expect(snack5000).toBeDefined();
+                fixture.detectChanges();
+
+                tick(firstTimeout);
+                const snackAfterTimeout5000 = getSnack();
+                expect(snackAfterTimeout5000).toBeNull();
+
+                const secondTimeout = 5000;
+                service[type](faker.lorem.paragraph(), firstTimeout);
```
sergiubologa: Here I think you forgot 5000 and should be 1000. Also, the second param should be `secondTimeout` instead of `firstTimeout `

vladjerca: 😬

sergiubologa:
Everything looks good from my side

projects/angular/components/ui-snackbar/src/ui-snackbar.component.ts
```diff
@@ -0,0 +1,177 @@
...
+     */
+    icon: string;
+    /**
+     * How long to remain on the screen
+     */
+    duration: number;
+    /**
+     * Aria label for screen-readers on close button
+     */
+    closeAriaLabel?: string;
```
anbalase: When using a snack error we include an error reference / id (e.g.: ` leet error message #(1337) `), should we extend the alert with an optional `messageRef`, and add to config a ref pattern ` #(${messageRef}) `

Overkill ?
@vladjerca - wdyt ?

vladjerca: That's specific internal logic that was introduced in order to test / report bugs.

I don't think this should be part of the alerting mecanism.

sergiubologa:
> LGTM - squash & merge

Done.

PR #44, opened 2019-07-18
chore: implement initial public api export checker

vladjerca:
- tries its best to determine if unnamed exports have annotations and throws an error if so
- checks to see if public decorated exports are hiding behind a barrel 
- command: `npm run check:exports`


vladjerca:
@sergiubologa, intially barrel checks weren't in place, so I didn't want to put it as part of the CI build.

I'll add it to the pipeline.

PR #65, opened 2019-12-12
feature / provide option for the grid to be without column header

rajeevuipath:
## Components
Grid: added key showHeaderButton to hide the column header for simple scenarios
Template : 
image


rajeevuipath:
> looks good!
> the only issue I'm having is that we're also hiding the progress bar, could you refactor the code/styles a bit so that it will remain visible even when not rendering the header please?

Updated :)

rajeevuipath:
Rebased, moved the progress bar outside of header row. and removed the redundant styles from scss files

PR #66, opened 2019-12-18
feat(grid): group by functionality with multiSort obervable

rajeevuipath:
## Grouping of rows with same entity property.
Column is having an input groupable, which if enabled then user will see a dropdown with list of columns which are groupable.  Groupable is supposed to be used with mutlisort$ observable in sort-manager which client can use to generate query. If not then user will not be able to sort other columns
At core, Column by which grid is grouped is the primary column by sorting is done, and ui-grid displays a "no data row" with entity's property's value or message returned by groupMessageCb function

## How to test
For any column add tags 
`groupable = "true"` and 
`groupMessageCb = "< optional custom function to override the message >"`

### Example
**Template**
```
<ui-grid-column [title]="'Task Status' "
                [groupable]="true"
                [groupMessageCb]="groupMessage"
                property="status"></ui-grid-column>
```
**Ts file**
```
public groupMessage(arg: string) {
    if (arg === TaskStatus.Unassigned || arg === TaskStatus.Pending) {
         return 'Action Needed';
    }
    return 'No Action Required';
}
```
image

### Sample application demo
https://uipath.enterprise.slack.com/files/WLLBB7PMH/FU2CE9JBV/zoom_1.mp4


projects/angular/components/ui-grid/src/body/ui-grid-column.directive.ts
```diff
@@ -162,6 +162,21 @@ export class UiGridColumnDirective<T> implements OnChanges, OnDestroy {
...
+    @Input()
+    public groupable = false;
+
+    /**
+     * Call back function which accepts the value of a property to return the desired message or property's value
+     * useful when entity's property is numeric or empty user wants it value to map to something else
+     *
+     */
+    @Input()
+    public groupMessageCb?: Function;
```
vladjerca: This should be part of the `intl` service provider.

vladjerca:
Also, we need this rebased and rechecked to make sure everything is fine with `ng9`

projects/angular/components/ui-grid/src/body/ui-grid-column.directive.ts
rajeevuipath: Are you sure? This is a callback function exposed to user to give the custom group header,  ultimately which is used to render `group row`. It is very specific to the grid's column. Localization of group header being rendered is more of the application's responsibility here.

vladjerca: The `intl` service can also be provided at the `component` level. It doesn't have to be global, grids that require grouping can extend the general `intl` provider and define the `groupMessage` method to that grid and provide it.

rajeevuipath: @vladjerca but it will not be specific to that column then and I won't able to compare strings in ui-grid.component.ts `this.groupMessage = (column.groupMessageCb && column.groupMessageCb(propertyValue)) || propertyValue;`.  Also I don't see the point, application using this will already be providing the appropriate strings to render. Let me know if I am missing something

vladjerca: I might be the one missing something, couldn't you simply do:

``` ts
this.groupMessage = intl.groupMessage(propertyValue));
``` 

```ts
// default implementation
public groupMessage = (propertyValue) => propertyValue;
```
I fail to see why we can't take this approach, by doing so any consumer will avoid injecting whatever translation provider used within the app in the component, which will allow `internationalization` to be tested separately from the actual component...

vladjerca:
Question: How does pagination work with this feature?

When I navigate to the next page, the response for the target page gets grouped? If so, that would be a weird behavior, it would mean that for every page I would have the group rows jumping around.

I honestly think that there should be paginators available for every group if what I assumed above is true. 

@toxik @anbalase thoughts?

projects/angular/components/ui-grid/src/body/ui-grid-column.directive.ts
rajeevuipath: I have pushed a ficup ccommit resolving this. Please have a look, if fixup commit is good, I will merge it

vladjerca:
@rajeevuipath Please rebase this branch and update the test definitions.

After that I'm personally ok with the changes.

@toxik @anbalase @mhagape Any other thoughts here?

projects/angular/components/ui-grid/src/ui-grid.component.ts
```diff
@@ -368,6 +369,23 @@ export class UiGridComponent<T extends IGridDataEntry> extends ResizableGrid<T>
...
+     * Array of groups and its state
+     *
+     */
+    public groups: { propertyValue: UiGridColumnDirective<T>['property'], message: any, collapsed?: boolean }[] = [];
+
+    /**
+     * group header message
+     *
+     */
+    public groupMessage: any;
```
toxik, line 387: shouldn't this be a string?
```suggestion
    public groupMessage: string;
```

rajeevuipath: Can't columns's property Value can be of type number, enum, etc other than string?

vladjerca: The value, sure... But the group`Messsage` should be a string

rajeevuipath: makes sense

projects/angular/components/ui-grid/src/ui-grid.component.html
```diff
@@ -34,6 +35,41 @@
...
         </ui-grid-search>
 
+        <ng-container *ngIf="groupableColumns.length">
+            <div class="ui-grid-dropdown-filter-container ui-grid-filter-option"
+                 data-cy="ui-grid-group-dropdown">
+                <button [matMenuTriggerFor]="groupOptions"
+                        mat-button
+                        type="button"
+                        class="ui-grid-dropdown-filter-button">
+                    <span class="ui-grid-dropdown-filter-title">Group By: </span>
```
anbalase, line 45: please add an `intl.groupByTitle` for this

toxik:
I took it for a spin and the sorting looks broken, have a look below (you need to unzip it first due to gh attachment limitations):

 - sorting looks disabled when you use the grouping feature, is that intended?
 - sorting has some glitches when it's reset, check below
 
[group-sort.mp4.zip](https://github.com/UiPath/angular-components/files/4474246/group-sort.mp4.zip)

rajeevuipath:
> I took it for a spin and the sorting looks broken, have a look below (you need to unzip it first due to gh attachment limitations):
> 
> * sorting (on other columns than grouped) looks ~disabled~ ignored when you use the grouping feature, is that intended?
> * sorting has some glitches when it's reset, check below
> 
> [group-sort.mp4.zip](https://github.com/UiPath/angular-components/files/4474246/group-sort.mp4.zip)

1. Groupable feature works on multiSort$ which emits all the columns clicked chronologically. Since orchestrator uses sort$, it won't work in the expected way. Any application using this need to subscribe to multisort$ for other columns to get sorted. 
Note: I have implemented in such a way that sorting will happen withing the group as I don't want to breakt
2. For your second issue, I tested the scenarios in a dummy application, and those glitches were not replicated there. It seems somewhere orchestrator is explicitly changing the column's sort direction and thus we are getting two columns with with non-empty sort direction. 

I have created a demo showing the use case/behvaior or sorting logic. Please see these to get the gist of the behavior. https://uipath.zoom.us/rec/share/3M1VLezN8GNIaYGRzQKDefADJ6fvX6a80ykZ_aAIzhzTtOmHr8QWfRiEbqp0WZ7T?startTime=1586969242000

rajeevuipath:
Closing this PR as of now, as its not needed in priority

PR #70, opened 2020-01-17
fix(suggest): remove tooltip after selection

toxik:
workaround for https://github.com/angular/components/issues/15648

this is a fix for the case when the tooltip gets 'hanged' after closing the virtual scrolled list

### DIFF
--- a/projects/angular/components/ui-suggest/src/ui-suggest.component.html
+++ b/projects/angular/components/ui-suggest/src/ui-suggest.component.html
@@ -124,7 +124,7 @@
             <ng-container *ngTemplateOutlet="customValueTemplate; context: { $implicit: 'down' }">
             </ng-container>
 
-            <ng-container *cdkVirtualFor="let item of renderItems; let index = index; trackBy: trackById">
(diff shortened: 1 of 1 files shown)


projects/angular/components/ui-suggest/src/ui-suggest.component.html
```diff
@@ -124,7 +124,7 @@
             <ng-container *ngTemplateOutlet="customValueTemplate; context: { $implicit: 'down' }">
             </ng-container>
 
-            <ng-container *cdkVirtualFor="let item of renderItems; let index = index; trackBy: trackById">
+            <ng-container *cdkVirtualFor="let item of renderItems; let index = index; trackBy: trackById; templateCacheSize: disableTooltip ? 20 : 0">
```
toxik: "magic" 20 value is taken from https://github.com/angular/components/blob/3e5e9db44ac74145304c9f427edbdb85c8291e91/src/cdk/scrolling/virtual-for-of.ts#L128

PR #73, opened 2020-02-06
chore: upgrade to ng @ 9

vladjerca:
TODO:

- [x] investigate test failures


projects/angular/components/ui-grid/src/ui-grid.component.html
```diff
@@ -240,7 +240,7 @@
         </div>
     </div>
     <div class="ui-grid-footer-container">
-        <mat-paginator *ngIf="!!footer"
+        <mat-paginator *ngIf="!!footer as footer"
```
sergiubologa: Why is `as` needed?

projects/angular/directives/ui-ng-let/src/ui-ng-let.directive.spec.ts
```diff
@@ -80,9 +80,7 @@ describe('Directive: UiClipboard', () => {
         `,
     })
     class NgLetDynamicFixtureComponent {
-        @ViewChild('child', {
-            static: false,
```
sergiubologa: So `static` is not required anymore?

projects/angular/components/ui-grid/src/ui-grid.component.html
vladjerca: It's nonsense, I've removed it. Ignore it I was really tired when porting 😬

projects/angular/directives/ui-ng-let/src/ui-ng-let.directive.spec.ts
vladjerca: No, it now defaults to `false`.

vladjerca:
@toxik @anbalase @sergiubologa Should be good to go.

projects/angular/directives/ui-virtual-scroll-range-loader/src/ui-virtual-scroll-range-loader.directive.spec.ts
```diff
@@ -236,7 +236,7 @@ describe('Directive: UiVirtualScrollRangeLoaderDirective', () => {
             }),
         );
 
-        it('should emit  RangeLoad without the last item index if Pula that is loaded',
```
sergiubologa: Should we search for `Pula` in the entire lib? 🤣 🤣 🤣

PR #74, opened 2020-02-13
fix(grid): announce page changes with unknown total

mhagape:


projects/angular/components/ui-grid/src/events/page-change-event.ts
```diff
@@ -0,0 +1,9 @@
+import { PageEvent } from '@angular/material/paginator';
+
+export type PageChangeEvent = Omit<PageEvent, 'length'> & {
```
anbalase: let's also export this type to the `public_api.ts` file

mhagape: Thanks, should be fixed now.

PR #78, opened 2020-02-18
fix(suggest): react to intl changes

vladjerca:


projects/angular/components/ui-suggest/src/ui-suggest.component.ts
```diff
@@ -568,6 +568,9 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
         );
 
         this.intl = this.intl || new UiSuggestIntl();
+        this.intl
+            .changes
```
anbalase: `takeUntil()` ? :)

If the viewRef has been destroyed detectChanges throws error, maybe use markForCheck?

vladjerca: Mark for check goes up the tree, checks need to be done at the component level. We just need to add the `takeUntil`

PR #86, opened 2020-04-01
feat(suggest): add support for minimum characters search

adrian-macuc:
TO-DO:
- [x] add unit tests
- [x] add hint message

### DIFF
--- a/projects/angular/components/ui-suggest/src/ui-suggest.component.ts
+++ b/projects/angular/components/ui-suggest/src/ui-suggest.component.ts
@@ -357,6 +357,7 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
      * Configure the `fetchStrategy` for requesting data using searchSourceFactory
      * `eager` - makes calls to searchSourceFactory onInit
      * `onOpen` - makes calls to searchSourceFactory onOpen
+     *
      */
     @Input()
     public set fetchStrategy(strategy: 'eager' | 'onOpen') {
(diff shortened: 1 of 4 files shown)


projects/angular/components/ui-suggest/src/ui-suggest.component.ts
```diff
@@ -357,13 +357,25 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
...
-    public set fetchStrategy(strategy: 'eager' | 'onOpen') {
+    public set fetchStrategy(strategy: 'eager' | 'onOpen' | 'minChars') {
         if (strategy === this._fetchStrategy$.value) { return; }
 
         this._fetchStrategy$.next(strategy);
     }
+
+    /**
+     * Configure the minimum number of characters that triggers the searchSourceFactory call
+     * Only works with `minChars` fetch strategy, otherwise will be ignored.
```
anbalase, line 373: Why do we need a new fetch strategy? Why not use the default value as `0`?
It should be simple to add this limit. The user shouldn't care about the value of the `fetchStrategy` input.

If `eager || onOpen` with `minChars === 0` it should work as before.

If the user specifies `[minChars]="3"` it should make the call only if `value.length >= minChars`, that's what I would expect.  
If you'd specified  `minChars` as limit, then you don't really want an `eager` behavior and that strategy becomes irrelevant. Yes `eager` strategy is not be meaningful and would act as `onOpen`, but I don't see an issue with this.

If at some point we'll add support to specify an `initialSearchValue` (let's say we have a search page with deep links and that's read from the query params) then `minChars` && `eager` might make sense to bind together...

adrian-macuc: I've added it because if you specify both the fetch strategy (either `onOpen` or `eager`) and `[minChars]=3`, it is not clear which one has priority, cause together they don't make sense at this point. If we choose not to add a new fetch strategy, the `[minChars]` will have priority as you said, but I wanted the fetch strategy to be the one. Also, `minChars` in fact **is** a fetch strategy and acts like one, but you somehow need to also specify the limit.



Also,

PR #92, opened 2020-05-14
feat(snackbar): add optional hardening when showing messages

toxik:
add support for injecting `UI_MAT_SNACK_BAR_DEFAULT_OPTIONS` which can enable `restrictHtml` - stripping the html tags and leaving only the text content in the shown message

### DIFF
--- a/projects/angular/components/ui-snackbar/src/ui-snackbar.component.spec.ts
+++ b/projects/angular/components/ui-snackbar/src/ui-snackbar.component.spec.ts
@@ -21,6 +21,8 @@ import {
     panelClass,
     SnackbarAction,
     SnackBarType,
+    UI_MAT_SNACK_BAR_DEFAULT_OPTIONS,
+    UiMatSnackBarConfig,
     UiSnackBarService,
 } from './ui-snackbar.component';
 import { UiSnackBarModule } from './ui-snackbar.module';
@@ -44,6 +46,7 @@ describe('Service: UiSnackBarService', () => {
     let service: UiSnackBarService;
     let overlayContainer: OverlayContainer;
     let fixture: ComponentFixture<SnackBarFixtureComponent>;
+    let securitySettings: UiMatSnackBarConfig;
 
     const getSnack = () =>
         overlayContainer
@@ -67,6 +70,10 @@ describe('Service: UiSnackBarService', () => {
     };
 
     beforeEach(async(() => {
+        securitySettings = {
+            restrictHtml: false,
+        };
+
         TestBed.configureTestingModule({
             imports: [
                 NoopAnimationsModule,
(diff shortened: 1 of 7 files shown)


projects/angular/components/ui-snackbar/src/ui-snackbar.component.spec.ts
```diff
@@ -305,4 +311,20 @@ describe('Service: UiSnackBarService', () => {
...
+        fixture = TestBed.createComponent(SnackBarFixtureComponent);
+        service = fixture.componentInstance.service;
+
+        service.show(`
+            <a id="injected-link" href="#some-link">a link text</a>
+            <img id="injected-image" src="invalid" onerror=callError()>
+            hello world
+        `);
+
+        const snack = getSnack();
```
mhagape, line 326: This test seems broken, if you inspect the DOM right after the `service.show` call you'll see that the anchor and image are actually rendered. We don't seem to use `securitySettings` in any way so I think this test isn't actually using the filtering mechanism that you added.

The test seems to pass because the rendered elements don't have the specified IDs—when fixing this, maybe we could make the selectors below more general (just `a` and `img`) to avoid false-positives in the future?

projects/angular/components/ui-snackbar/src/ui-snackbar.component.ts
```diff
@@ -169,6 +181,13 @@ export class UiSnackBarService {
         })
 
     private _alert(type: SnackBarType, options: ISnackBarAlert) {
+        if (this._additionalOptions?.restrictHtml) {
+            const span = document.createElement('span');
+            span.innerHTML = this._sanitizer.sanitize(SecurityContext.HTML, options.message) || '';
+            options.message = span.innerText;
+            span.remove();
```
mhagape, line 188: We should [fix the test](https://github.com/UiPath/angular-components/pull/92/files#r426609967) then take a closer look here: as we're using the `innerText`, we're still open to relatively simple attack vectors e.g. `service.show('&lt;a href="uipath.com"&gt;Click me&lt;/a&gt;')` will result in the `Click me` anchor being rendered.
image

projects/angular/components/ui-snackbar/src/ui-snackbar.component.spec.ts
toxik: good catch, thanks!

projects/angular/components/ui-snackbar/src/ui-snackbar.component.ts
toxik: VERY good catch; the initial fix would have made the attach a bit counter intuitive but a simple lookup on the source would have revealed this.
I applied your suggestion with `innerText = message` and all is fine now.
Thank you!

toxik: diff here https://github.com/UiPath/angular-components/compare/f0ad4d8827b737827d628911a8e7f2e938794a68..3079c2dbcf822af3f5d6c6ea3f448a5cbe9bf528

PR #93, opened 2020-06-12
feat(a11y): add auto aria-label support; fix(grid): a11y column toggle

anbalase:


projects/angular/a11y/src/ui-auto-accessible-label/ui-auto-accessible-label.directive.ts
```diff
@@ -0,0 +1,40 @@
...
+    HostBinding,
+    Input,
+    Optional,
+} from '@angular/core';
+import { MatTooltip } from '@angular/material/tooltip';
+
+export const DISABLE_AUTO_ACCESSIBLE_LABEL_ATTRIBUTE = 'disable-auto-accessible-label';
+
+// tslint:disable: max-line-length
+export const MAT_TOOLTIP_MISSING_WARNING = `
```
anbalase, line 13: ❤️

projects/angular/components/ui-grid/src/components/ui-grid-toggle-columns/ui-grid-toggle-columns.component.html
```diff
@@ -19,14 +19,19 @@
                 <mat-icon>view_column</mat-icon> {{ toggleTitle }}
             </span>
 
-            <button *ngIf="dirty"
-                    [matTooltip]="resetToDefaults"
-                    (click)="reset()"
-                    type="button"
-                    color="primary"
-                    mat-button>
+            <a #resetBtn
```
mhagape, line 22: Can't we keep this as a `button`? We'd get `enter` and `space` support for free. If not, I think we should add `[role=button]` and remove the `type` attribute.

projects/angular/components/ui-grid/src/ui-grid.component.spec.ts
```diff
@@ -1741,6 +1747,218 @@ describe('Component: UiGrid', () => {
...
+
+                it('should render an open select menu', () => {
+                    const panel = fixture.debugElement.query(By.css('.ui-grid-toggle-panel'));
+
+                    expect(panel).toBeDefined();
+                });
+
+                it('should render reset button if option is changed', fakeAsync(
+                    () => {
+                        const firstOption = fixture.debugElement
```
mhagape, line 1772: Note (to self?): we should look into [component harnesses](https://material.angular.io/cdk/test-harnesses/overview).

projects/angular/components/ui-grid/src/components/ui-grid-toggle-columns/ui-grid-toggle-columns.component.html
anbalase: Well there's a parent label there. 😄 
And having a button inside a label had an unexpected side-effect.
Clicking anywhere on the label `<mat-icon>view_column</mat-icon> {{ toggleTitle }}` triggered a click event on the button as well, due to the fact that the button was inside the label.
Instead of `preventDefault` on the <span> I went with changing this to an anchor.

projects/angular/components/ui-grid/src/ui-grid.component.spec.ts
anbalase: Yes, it's a refactor in itself 😄

PR #96, opened 2020-08-24
feat: upgrade to Angular 10

sergiubologa:
- [x] upgrade Angular core and cli
- [x] upgrade Angular Material
- [x] removed `entryComponents` which are not mandatory anymore with the Ivy compiler and runtime
- [x] fix(virtual-scroll-range-loader): filter range when empty or touched
- [x] fix: change import sorter to match tslint order ruleset
- [x] chore: define recommended extensions
- [x] chore: format files


projects/angular/components/ui-grid/src/ui-grid.intl.ts
```diff
@@ -9,6 +12,7 @@ import { IDropdownOption } from './filters/ui-grid-dropdown-filter.directive';
  *
  * @export
  */
+@Injectable()
```
sergiubologa, line 15: Adding `@Injectable()` decorator to classes that implement Angular lifecycle methods will become mandatory in the next major version of Angular, therefore I think it's useful to add it now. 

I think the change should be fine and not break anything, but please @UiPath/team-green test it in Orchestrator to make sure the Grid is translated properly in integration.

sergiubologa:
@UiPath/team-green 

Not really sure how to solve the failing tests...

Seems like there's a change in behavior for the `renderedRangeStream` stream from CdkVirtualScrollViewport which, for the case bellow, on version 9 emits the range `0-5` and on version 10 it returns the range `0-0`.

I'm not sure if it's a bug in the cdk or it was a bug and they've fixed it in version 10. Can anyone with some ui-suggest knowledge please help? I don't know how should behave in this case...

image

toxik (commented):
maybe we should also bump the version `9.0.7` to `10.0.0` ?

adrian-macuc (approved):

PR #103, opened 2020-10-09
fix(grid): multiple row selection with shift fix

CalinaCristian:
When shift key is used while multi selecting, the checkbox appeared as unselected even though in the model it was selected.
Performance improvements for shift selection algorithm (instead of deselecting everything and selecting again, we calculate the rows to be selected and the rows to be deselected excepting the ones that already are in correct state).

### DIFF
--- a/projects/angular/components/ui-grid/src/ui-grid.component.ts
+++ b/projects/angular/components/ui-grid/src/ui-grid.component.ts
@@ -8,7 +8,6 @@ import {
     ElementRef,
     EventEmitter,
     HostBinding,
-    HostListener,
     Input,
     NgZone,
     OnChanges,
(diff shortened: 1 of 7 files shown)


projects/angular/components/ui-grid/src/ui-grid.component.ts
```diff
@@ -602,16 +601,24 @@ export class UiGridComponent<T extends IGridDataEntry> extends ResizableGrid<T>
...
-         * If min = max, the checkbox will be deselected as a consequence of clicking
-         * and will remain rendered as unchecked, even though we selected it
-         * to prevent this invalid render state, we check for changes after deselecting the rows
+         * To be consistent with the browser, if we click on a row
+         * that was already selected, we unselect it, sync with DOM (detectChanges),
+         * then we select it again (it's included in rowsForSelection).
          */
+        if (this.selectionManager.isSelected(entry)) {
+            this.selectionManager.deselect(entry);
+            this._cd.detectChanges();
```
anbalase, line 616: why is this needed? same call is done on line :622

anbalase (approved):
lgtm

PR #105, opened 2020-10-22
fix(a11y): grid announce

anbalase:
- [x] feat(**grid**) differentiate between user sort and programmatic sort events
- [x] fix(**grid**) a11y: announce only user `sort` events
- [x] fix(**grid**) a11y: translatable aria-labels for checkboxes 
- [x] merge #106

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+# v9.1.0 (2020-10-21)
+* **grid** differentiate between user sort and programmatic sort events
+* **grid** a11y: announce only user `sort` events
+* **grid** a11y: expose translateable aria-label for checkboxes
+* **grid** added `matTooltip` for checkboxes
+* **suggest** a11y: fixes to title, specify `role` attributes for list
+* **suggest** a11y: announce current `option` on open
+
 # v9.0.8 (2020-10-15)
 * **grid** fix multiple row selection with shift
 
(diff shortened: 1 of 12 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# v9.1.0 (2020-10-21)
+* **grid** differentiate between user sort and programmatic sort events
+* **grid** announce only user sort events
```
CalinaCristian, line 3: this should be the last commit, with every change added

CalinaCristian (approved):
LGTM:

Few mentions:
- On NVDA it might have some issues (it reads weirdly the checkboxes -- for select on each row it reads Select All as well, but it doesn't reproduce on JAWS). It also reads HEADER row as row 1, then first row of data is row 2, etc. It kinda mixes the row count, saying row 0 and row 2 as well for the same row (counting the header as row 1 or not counting it at all ). On JAWS for instance it starts counting from 0 when selecting a row of data.

The red border is for extra text that NVDA reads and JAWS doesn't. The blue one is what they both read.

image



- On JAWS, when you're tabbing through the header row of the grid, if you go from an input or a checkbox (like select all or actions) to a column title it will read the column number as well -- like on NVDA, but if you're going from a column title to another it won't read the column number anymore (unlike in the above picture, for NVDA). It will read the header as row 1, and then when it goes through the data it starts from row 0 again.

Those might be bugs of NVDA/JAWS that we can live with (or not), but the issues with the announcer on first load were fixed by this PR.

PR #122, opened 2021-03-08
fix(collapsible-filters) hide when filters not visible, disable when grid disabled

CalinaCristian:


projects/angular/components/ui-grid/src/ui-grid.component.ts
```diff
@@ -605,6 +613,19 @@ export class UiGridComponent<T extends IGridDataEntry> extends ResizableGrid<T>
             this.rendered.next();
         });
 
+        combineLatest(this.columns.map((column: UiGridColumnDirective<T>) =>
+            column.dropdown?.visible$ || column.searchableDropdown?.visible$ || of(false),
+        )).pipe(
+            takeUntil(this._destroyed$),
```
anbalase, line 619: Lets map in a pipe to `hasVisible` and add a `distinctUntilChanged()` as well.

anbalase: Not a big fan of having to maintain another inner subscription. 
Can we directly assign our composed stream `hasAnyFiltersVisible$` and have it start out with false? 
Or, could we add it as a side-effect of an existing subscription?

projects/angular/components/ui-grid/src/ui-grid.component.ts
```diff
@@ -605,6 +613,19 @@ export class UiGridComponent<T extends IGridDataEntry> extends ResizableGrid<T>
...
+        ).subscribe(areVisible => {
+            const hasVisible = areVisible.some(visible => visible);
+
+            if (!hasVisible) {
```
anbalase, line 623: What happens if I started out with an input `[showFilters]=true` then I get successive emissions: false, true? 

Does `this.showFilters` stay false? 
Seems unexpected to have this side-effect. 
I'd recommend subscribing to this stream as well in the template rather than overriding the input.

anbalase (changes requested):

anbalase (approved):
lgtm

anbalase (approved):

PR #124, opened 2021-03-09
feat(grid): update no content template

robertandreimoldoveanu:


projects/angular/components/ui-grid/src/ui-grid.component.html
```diff
@@ -374,9 +374,10 @@
 <ng-template #defaultNoData>
     <div *ngIf="!dataManager.pristine"
          class="ui-grid-row ui-grid-no-data-container"
+         [class.ui-grid-no-content-available]="!header?.searchValue && !(filterManager.activeCount$ | async)"
```
CalinaCristian, line 377: subscribe only once with a ngLet?

CalinaCristian (commented):
Maybe add some tests as well for a grid with noDataMessage passed?

projects/angular/components/ui-grid/src/ui-grid.component.html
vladjerca: `ngLet`? https://github.com/UiPath/angular-components/blob/master/projects/angular/directives/ui-ng-let/src/ui-ng-let.directive.ts

anbalase, line 377: How are we going to accomplish a different layout for no content (with no filters or search)? It's more than just text alignment, we also need another icon. Maybe we should branch out to a different template based on `useAlternateDesign`?

projects/angular/components/ui-grid/src/ui-grid.intl.ts
```diff
@@ -102,7 +97,11 @@ export class UiGridIntl implements OnDestroy {
      *
      */
     public descending = 'descending';
-
+    /**
+     * No data row message.
+     *
+     */
+    public noDataMessage = (_searchValue?: string, _activeFilters?: number | null) => 'No data is available.';
```
anbalase, line 104: This will be a breaking change, let's use another one for an alternate message.

projects/angular/components/ui-grid/src/ui-grid.component.html
robertandreimoldoveanu: I was thinking of adding the icon from CSS in Orchestrator (attaching SVG on a pseudo element).
I find that this way the grid is more consistent - it's always displaying text - and also allows room for change depending on the scenario

CalinaCristian (dismissed):

anbalase (changes requested):

projects/angular/components/ui-grid/src/ui-grid.intl.ts
```diff
@@ -102,7 +102,11 @@ export class UiGridIntl implements OnDestroy {
      *
      */
     public descending = 'descending';
-
+    /**
+     * No data row message alternative function.
+     *
+     */
+    public noDataMessageAlternative = (_searchValue?: string, _activeFilters?: number | null) => 'No data is available.';
```
anbalase, line 109: Looks a bit wrong to declare these arguments with leading `_`.
Also, shouldn't we have tests for the grid for this new logic?
```suggestion
    public noDataMessageAlternative = (searchValue?: string, activeFilters?: number | null) => `No data is available ${searchValue ? ` for ${searchValue}${activeFilters? ' for applied filters': ''}`}.`;
```

projects/angular/components/ui-grid/src/ui-grid.component.scss
```diff
@@ -183,6 +183,17 @@ ui-grid {
                         order: 1;
                     }
                 }
+
+                .ui-grid-no-data-container {
+                    height: 4 * $ui-grid-row-height;
```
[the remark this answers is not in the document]
robertandreimoldoveanu: Might be, but if they have a `pageSize` too big, the centered message will be below the fold and not visible. I think we should ask design.

anbalase (approved):

PR #135, opened 2021-03-19
fix: a11y grid 

robertandreimoldoveanu:
- [x] add cell headers association - assure row headers
- [x] reveal changes in header when a selection is made and action buttons are available
  - the user is announced of the change and he can also press `Shift + ALT + Arrow Up` to navigate straight to the newly revealed actions
- [x] add enter and space to event listeners for drag and drop directive to ensure the file browsing can be triggered by keyboard

### DIFF
--- a/projects/angular/components/ui-grid/src/body/ui-grid-column.directive.spec.ts
+++ b/projects/angular/components/ui-grid/src/body/ui-grid-column.directive.spec.ts
@@ -49,7 +49,7 @@ class TestFixtureComponent {
     public method?: string;
     public sort?: 'asc' | 'desc' | '';
     public refetch?: boolean;
-    public primary?: boolean;
+    public primary = false;
     public minWidth?: number;
 }
 
@@ -92,6 +92,9 @@ describe('Component: UiGridColumn', () => {
             it('should have visible set to false', () => {
                 expect(column.visible).toBeFalse();
             });
+            it('should have primary set to false', () => {
+                expect(column.primary).toBeFalse();
+            });
 
             it('should have all properties undefined', () => {
                 expect(column.width).toBeNaN();
@@ -103,7 +106,6 @@ describe('Component: UiGridColumn', () => {
                 expect(column.property).toBeUndefined();
                 expect(column.method).toBeUndefined();
                 expect(column.sort).toBeUndefined();
-                expect(column.primary).toBeUndefined();
                 expect(column.minWidth).toBeUndefined();
             });
         });
--- a/projects/angular/components/ui-grid/src/body/ui-grid-column.directive.ts
+++ b/projects/angular/components/ui-grid/src/body/ui-grid-column.directive.ts
@@ -34,7 +34,7 @@ const ARIA_SORT_MAP: Record<SortDirection, string> = {
  * @ignore
  */
 const REACTIVE_INPUT_LIST: (keyof UiGridColumnDirective<{}>)[]
-    = ['sort', 'visible', 'title'];
+    = ['sort', 'visible', 'title', 'primary'];
 
 /**
  * The grid column definition directive.
@@ -144,7 +144,17 @@ export class UiGridColumnDirective<T> implements OnChanges, OnDestroy {
      *
      */
     @Input()
-    public primary = false;
+    public get primary() {
+        return this._primary;
+    }
+    public set primary(primary: boolean) {
+        if (primary === this._primary) { return; }
+        this._primary = !!primary;
+
+        this.change$.next({
+            primary: new SimpleChange(!primary, primary, false),
+        });
+    }
 
     /**
      * If the column can have visibility toggled.
@@ -219,6 +229,7 @@ export class UiGridColumnDirective<T> implements OnChanges, OnDestroy {
 
     private _width = NaN;
     private _visible = true;
+    private _primary = false;
 
     /**
      * @ignore
--- a/projects/angular/components/ui-grid/src/managers/selection-manager.ts
+++ b/projects/angular/components/ui-grid/src/managers/selection-manager.ts
@@ -1,6 +1,10 @@
 import cloneDeep from 'lodash-es/cloneDeep';
 import differenceBy from 'lodash-es/differenceBy';
-import { Subject } from 'rxjs';
+import {
+    BehaviorSubject,
+    Subject,
+} from 'rxjs';
+import { distinctUntilChanged } from 'rxjs/operators';
 
 import {
     SelectionChange,
@@ -33,6 +37,10 @@ export class SelectionManager<T extends IGridDataEntry> {
     }
 
     public changed$: Subject<SelectionChange<T>> = new Subject();
+
+    private _hasValue$ = new BehaviorSubject(false);
+
+    public hasValue$ = this._hasValue$.pipe(distinctUntilChanged());
 
     private _selection = new Map<number | string, T>();
 
@@ -93,6 +101,7 @@ export class SelectionManager<T extends IGridDataEntry> {
     private _updateState = (predicate: (value: T) => void, values: T[]) => {
         values.forEach(predicate);
         this._emitChangeEvent();
+        this._hasValue$.next(this.hasValue());
     }
 
     private _emitChangeEvent() {
--- a/projects/angular/components/ui-grid/src/ui-grid.component.html
+++ b/projects/angular/components/ui-grid/src/ui-grid.component.html
@@ -271,6 +271,7 @@
                 </mat-checkbox>
             </div>
             <div *ngIf="!!actions"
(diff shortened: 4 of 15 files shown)


projects/angular/components/ui-grid/src/ui-grid.component.spec.ts
```diff
@@ -255,6 +256,13 @@ describe('Component: UiGrid', () => {
                         expect(nestedDateCell.nativeElement.innerText).toContain(dataEntry.myObj.myObjDate.toString());
                     });
                 });
+
+                it('should render first column cells as row headers', () => {
```
anbalase, line 260: Let's write more meaningful tests, let's check the following scenarios pls:
3 columns, no primary => should mark the first column as a row header.
3 columns, 2nd & 3rd are primary => should mark only the 1st primary as a row header.

projects/angular/components/ui-grid/src/ui-grid.component.ts
```diff
@@ -781,4 +797,18 @@ export class UiGridComponent<T extends IGridDataEntry> extends ResizableGrid<T>
         }
         return `${this.selectionManager.isSelected(row) ? 'deselect' : 'select'} row ${this.dataManager.indexOf(row)}`;
     }
+
+    public isRowHeader(columns?: UiGridColumnDirective<T>[], index?: number) {
```
anbalase, line 801: When can this be called with no arguments?
```suggestion
    public isRowHeader(columns: UiGridColumnDirective<T>[], index: number) {
```

projects/angular/components/ui-grid/src/ui-grid.component.html
```diff
@@ -294,28 +297,30 @@
...
-                    </ng-container>
+        <ng-container *ngLet="visible$ | async as visibleColumns">
+            <ng-container *ngFor="let column of visibleColumns; let index = index;">
+                <div [class.ui-grid-primary]="column.primary"
+                     [class.ui-grid-secondary]="!column.primary"
+                     [class.ui-grid-state-resizing]="column === resizeManager.current?.column"
+                     [attr.data-property]="column.property"
+                     [attr.data-identifier]="column.identifier"
+                     [style.width.%]="column.width"
+                     [attr.role]="isRowHeader(visibleColumns, index) ? 'rowheader' : 'gridcell'"
```
anbalase, line 308: I'd recommend not having method calls that set attributes, especially in a *ngFor for each cell...
Let's compose a stream for this added state, starting from visible$ to optimize this.

anbalase (changes requested):

projects/angular/components/ui-grid/src/managers/selection-manager.ts
```diff
@@ -34,6 +37,8 @@ export class SelectionManager<T extends IGridDataEntry> {
 
     public changed$: Subject<SelectionChange<T>> = new Subject();
 
+    public hasValue$: BehaviorSubject<boolean> = new BehaviorSubject<boolean>(false);
```
anbalase, line 40: Let's expose this as an Observable with `distictintUntilChanged()`
```suggestion
    private _hasValue$ = new BehaviorSubject<boolean>(false);
    public hasValue$ = this._hasValue$.pipe(distictintUntilChanged());
```

projects/angular/components/ui-grid/src/ui-grid.component.ts
```diff
@@ -496,6 +505,26 @@ export class UiGridComponent<T extends IGridDataEntry> extends ResizableGrid<T>
         tap(compensationWidth => this.scrollCompensationWidth = compensationWidth),
     );
 
+    public hasSelection$ = this.selectionManager.hasValue$.pipe(
+        distinctUntilChanged(),
```
anbalase, line 509:
```suggestion
```

projects/angular/components/ui-grid/src/ui-grid.component.ts
```diff
@@ -496,6 +505,26 @@ export class UiGridComponent<T extends IGridDataEntry> extends ResizableGrid<T>
...
+        tap((hasSelection) => {
+            if (hasSelection && !!this.header?.actionButtons?.length) {
```
anbalase, line 511: if we have 2 subscriptions then we will announce twice, this should be avoided.

projects/angular/components/ui-grid/src/body/ui-grid-column.directive.spec.ts
```diff
@@ -92,6 +92,9 @@ describe('Component: UiGridColumn', () => {
             it('should have visible set to false', () => {
                 expect(column.visible).toBeFalse();
             });
+            it('should have primary set to false', () => {
```
anbalase, line 95: why did we change this test? 😄

projects/angular/components/ui-grid/src/ui-grid.component.spec.ts
```diff
@@ -255,6 +255,27 @@ describe('Component: UiGrid', () => {
                         expect(nestedDateCell.nativeElement.innerText).toContain(dataEntry.myObj.myObjDate.toString());
                     });
                 });
+
+                it('should mark first column cells as row headers when no primary column is set', () => {
+                    fixture.detectChanges();
+                    const firstRow = fixture.debugElement.queryAll(By.css('.ui-grid-row'))[0];
```
anbalase, line 261:
```suggestion
                    const firstRow = fixture.debugElement.query(By.css('.ui-grid-row'));
```

projects/angular/components/ui-grid/src/ui-grid.component.spec.ts
```diff
@@ -255,6 +255,27 @@ describe('Component: UiGrid', () => {
...
+                    expect(secondCell.nativeElement.getAttribute('role')).toBe('gridcell');
+                    expect(thirdCell.nativeElement.getAttribute('role')).toBe('gridcell');
+                });
+
+                it('should mark only one column cell as row header when there are multiple primary columns', fakeAsync(() => {
+                    grid.columns.forEach((col, idx) => col.primary = idx > 0);
+                    fixture.detectChanges();
+                    tick(GRID_COLUMN_CHANGE_DELAY);
+                    fixture.detectChanges();
+                    const firstRow = fixture.debugElement.queryAll(By.css('.ui-grid-row'))[0];
```
anbalase, line 273:
```suggestion
                    const firstRow = fixture.debugElement.query(By.css('.ui-grid-row'));
```

projects/angular/components/ui-grid/src/ui-grid.component.html
```diff
@@ -47,7 +47,8 @@
         </ng-container>
     </div>
 
-    <div *ngIf="selectionManager.hasValue()"
+    <div #gridActionButtons
+         *ngIf="hasSelection$ | async "
```
anbalase, line 51: There are other places where we can switch to this stream, could we update all the places where `selectionManager.hasValue()` is being used?
```suggestion
         *ngIf="hasSelection$ | async"
```

anbalase (changes requested):

projects/angular/components/ui-grid/src/body/ui-grid-column.directive.spec.ts
robertandreimoldoveanu: Before, the primary value was unreactive and the `change$` stream didn't emit when the primary input was changed, thus causing problems with the `ui-grid` tests. I changed how the property behaves and also it's default value, so that's why I changed the test as well.

projects/angular/components/ui-grid/src/body/ui-grid-column.directive.ts
```diff
@@ -217,8 +227,16 @@ export class UiGridColumnDirective<T> implements OnChanges, OnDestroy {
      */
     public change$ = new Subject<SimpleChanges>();
 
+    /**
+     * The a11y role that cells under this columns should have
+     *
+     * @ignore
+     */
+    public role?: string;
```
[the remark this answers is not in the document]
robertandreimoldoveanu: The type-checker throws errors otherwise :( 
It's not a side effect, a map operator that sets the property when rendering the visible columns. So right now, it is always set where we need it to be set. 
It's true not all consumers have it set, though. Perhaps we could add a default value instead of it being optional, but given that it's not used somewhere else, it wouldn't have any benefit. This was added to assure a better accessibility, I don't see how it could be used in other places.
Perhaps we could extend that class and create a new type, would that work better?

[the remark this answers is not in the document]
anbalase: Could we just define a new type and compose a new object where we consume the stream?

anbalase (approved):

PR #148, opened 2021-04-26
chore: add automatic changelog & version bump & tag

CalinaCristian:
Implemented `version-bump` script which follows these principles:

It takes 1 required argument and 1 optional one:

First is the type of the bump (`patch`, `minor`, `major`, `stable`, `pre` or `fix-changelog-order`).
- `stable` means that if you have version `v11.0.0-rc.3` and you do `node version-bump.js stable`, it will remove the `-rc.3`. 
- `pre` can be used with `patch`, `minor`, `major`,  to bump that version and add -rc.0 to it, or without this second argument and it will bump the version of rc (from `rc.0` to `rc.1` for example)
- `patch`, `minor`, `major` work as you'd expect them to.
- `fix-changelog-order` is used to regenerate the changelog based on it's content, but sorted by the date & tag version. If you have two branches that add to the changelog, and you merge the second into the first, there will be conflicts on changelog. You can accept both, and run this script.

There are scripts added to `package.json` that handle all of those cases.

After bumping the version, it also tags it (without pushing it, so you'll have to `git push --tags` manually)

Then, it reads all the commits from the previous tag, and formats them to the current changelog standard, prepending them to the `CHANGELOG.md` file. If it doesn't find commits between the previous tag and this one, it will just add the title. It also ignores all the `chore: bump version` commits.

## TODO:
Agree on a conventional pattern for commit messages on breaking changes, so we can use that as a regex to put the **BREAKING CHANGE** as well.


anbalase (approved):
lgtm, did not test

PR #157, opened 2021-05-26
feature/a11y-testing

robertandreimoldoveanu:


package.json
```diff
@@ -119,6 +119,7 @@
     "gh-pages": "2.0.1",
     "glob": "^7.1.4",
     "husky": "2.3.0",
+    "jasmine-axe": "^1.0.0",
```
anbalase, line 122: test conversation

anbalase (changes requested):
Could we add the same basic test for all components?Let's target next release with this.
Pls don't forget to do the version bump so that it will be ready for next release.

anbalase (approved):

CalinaCristian (approved):

PR #181, opened 2021-11-09
🚧feat(shortcut-manager): implement functionality

robertandreimoldoveanu:


sergiubologa:
FYI: in du-lib, there's a similar functionality, we call it `hotkeys`. Not sure if it fits all your needs, but I would recommend giving it a try. It can be found [here](https://github.com/UiPath/Activities/tree/develop/IntelligentOCR/DU.Web/projects/du-lib/services/src/hotkeys).

Unfortunately, it's not documented yet, but you can find usage examples in Validation Station and Classification Station. One example would be [here](https://github.com/UiPath/Activities/blob/develop/IntelligentOCR/DU.Web/projects/du-lib/components/ui-validation-station/src/services/hotkeys/vs-hotkeys.service.ts).

PR #183, opened 2021-12-06
Chore/Upgrade to Angular 13

alexdudescu:


projects/angular/test.ts
```diff
@@ -60,6 +60,9 @@ setSpecFn(it, fit, xit);
 getTestBed().initTestEnvironment(
     BrowserDynamicTestingModule,
     platformBrowserDynamicTesting(),
+    {
+        teardown: { destroyAfterEach: false },
```
anbalase, line 64: could we enable this?

package.json
```diff
@@ -1,167 +1,167 @@
...
+    "clipboard": "2.0.8",
+    "core-js": "^3.19.3",
+    "lodash-es": "4.17.21",
+    "moment": "2.29.1",
+    "moment-timezone": "0.5.33",
+    "object-hash": "2.2.0",
+    "rxjs": "7.4.0",
+    "scroll-into-view-if-needed": "2.2.28",
+    "tslib": "^2.3.1",
+    "zone.js": "~0.11.4"
```
anbalase, line 91: let's check if these can be slightly bumped as well with any available patches

projects/angular/test.ts
alexdudescu: 👍🏻

anbalase (approved):

PR #199, opened 2022-01-19
fix: suggest do not select on space

anbalase:


CalinaCristian (approved):

projects/angular/components/ui-suggest/src/ui-suggest.component.html
```diff
@@ -108,8 +108,7 @@
                        (keydown.arrowdown)="navigate(1, $event)"
                        (keyup.enter)="preventDefault($event);"
                        (keydown.enter)="preventDefault($event); isOpen ? setSelectedItem() : open();"
-                       (keyup.space)="isOpen ? setSelectedItem() : open();"
-                       (keydown.space)="preventDefault($event)"
+                       (keyup.space)="!isOpen && open();"
```
alexdudescu, line 111: nit: since there's no other following logic, can't we drop the `;` ? 🤔 
```suggestion
                       (keyup.space)="!isOpen && open()"
```

alexdudescu (approved):

PR #206, opened 2022-01-28
fix(suggest): prevent fetch after select

alexdudescu:
When selecting an item after search, the items value was updated,
thus causing an extra empty fetch

### DIFF
--- a/projects/angular/components/ui-suggest/src/ui-suggest.component.ts
+++ b/projects/angular/components/ui-suggest/src/ui-suggest.component.ts
@@ -16,6 +16,7 @@ import {
     filter,
     finalize,
     map,
+    pairwise,
     retry,
     startWith,
     switchMap,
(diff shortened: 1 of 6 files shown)


robertandreimoldoveanu (approved):

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
```diff
@@ -651,6 +652,10 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
             ),
             this._inputChange$,
         ).pipe(
+            startWith(''),
+            pairwise(),
+            filter(([prev, current]) => !(prev !== '' && current === '')),
```
anbalase, line 657: not 100% about this. 
If we wish to trigger a hard reset `this._reset$` this should always pass. Right now it seems to be filtered

robertandreimoldoveanu: Why would we want to fetch new data after a reset?

anbalase: also, we seem to filter out actual queries here.

anbalase (changes requested):
We're solving this issue on the other PR.

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
anbalase: @robertandreimoldoveanu That's the point of the reset stream to trigger a new fetch. UiSuggest can be used as a grid filter, there are case where we call this mechanism.

PR #212, opened 2022-02-17
feat(grid): add filters empty state input

CatalinB7:
Create optional input `emptyValue` for `UiGridDropdownFilterDirective` in order to clearly count the applied filters of the grid

### DIFF
--- a/projects/angular/components/ui-grid/src/ui-grid.component.html
+++ b/projects/angular/components/ui-grid/src/ui-grid.component.html
@@ -131,8 +131,9 @@
                          tabindex="0">
                         <div (click)="sortManager.changeSort(column)"
                              class="ui-grid-header-title">
-                            <p [class.ui-grid-header-title-filtered]="column.dropdown?.value != null ||
-                                                                 column.searchableDropdown?.value != null"
+                            <p [class.ui-grid-header-title-filtered]="(column.dropdown?.value != null &&
+                                    column.dropdown!.value!.value !== column.dropdown!.emptyStateValue) ||
+                                    column.searchableDropdown?.value != null"
                                [matTooltip]="column.title ?? ''"
                                [matTooltipDisabled]="resizeManager.isResizing">{{ column.title }}</p>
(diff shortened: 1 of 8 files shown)


projects/angular/components/ui-grid/src/ui-grid.component.html
```diff
@@ -131,7 +131,7 @@
                          tabindex="0">
                         <div (click)="sortManager.changeSort(column)"
                              class="ui-grid-header-title">
-                            <p [class.ui-grid-header-title-filtered]="column.dropdown?.value != null ||
+                            <p [class.ui-grid-header-title-filtered]="(column.dropdown?.value != null && column.dropdown?.value?.value !== column.dropdown?.emptyStateValue) ||
```
anbalase, line 134:
```suggestion
                            <p [class.ui-grid-header-title-filtered]="(column.dropdown?.value != null && column.dropdown!.value!.value !== column.dropdown!.emptyStateValue) ||
```

projects/angular/components/ui-grid/src/ui-grid.component.spec.ts
```diff
@@ -2805,6 +2813,54 @@ describe('Component: UiGrid', () => {
                     expect(collapisbleFiltersToggle).toBeFalsy();
                 });
 
+                it('should not count applied default filter when there is not empty filter state', () => {
```
CalinaCristian, line 2816: this should (and is) not be the behavior.. you apply the count on default filters every time.
As we spoke, this seems to be a broken setup for this test.

Maybe create a different Fixture component for those specific tests.

CalinaCristian (changes requested):
Fix the unit tests, otherwise it seems fine 🚀

CalinaCristian (dismissed):

CalinaCristian:
Don't forget about the chore: bump version commit & tag

alexdudescu (approved):

alexdudescu (approved):

PR #220, opened 2022-03-16
chore(grid): change column icon and divider height

CatalinB7:
1. set divider height 32px
2. change icon of column button with `table_chart` icon from material
3. add an arrow for column button


anbalase (dismissed):

anbalase:
all green https://uipath.visualstudio.com/angular-components/_build/results?buildId=2470593&view=results

PR #233, opened 2022-06-02
Feat/custom header items slot

alexdudescu:
Component: Suggest 

Have a slot at the top of the dropdown list to display a list of custom elements.

Todo: 
- [x] display custom items and separate with divider
- [x] correctly manage activeIndex when option is used
- [x] update tests

### DIFF
--- a/projects/angular/components/ui-suggest/src/ui-suggest.component.html
+++ b/projects/angular/components/ui-suggest/src/ui-suggest.component.html
@@ -231,52 +231,34 @@
             <ng-container *ngTemplateOutlet="customValueTemplate; context: { $implicit: 'down' }">
             </ng-container>
 
-            <ng-container
-                          *cdkVirtualFor="let item of renderItems; let index = index; trackBy: trackById; templateCacheSize: disableTooltip ? 20 : 0">
-                <mat-list-item *ngLet="drillDown && !!item?.expandable as itemExpandable"
-                               [class.active]="index === activeIndex"
-                               [class.is-loading]="item.loading !== VirtualScrollItemStatus.loaded"
-                               [class.readonly]="item.loading !== VirtualScrollItemStatus.loaded"
-                               [class.selected]="!!item &&
-                                                !multiple &&
-                                                isItemSelected(item)"
-                               [class.is-expandable]="itemExpandable"
-                               [style.height.px]="!!item ? itemSize : 0"
-                               (click)="preventDefault($event);
-                                        updateValue(item, !multiple, true);"
-                               [attr.role]="'option'"
-                               matRipple>
-                    <ng-container *ngIf="!!item">
-                        <ng-container *ngIf="itemTemplate; else defaultItem">
-                            <ng-container *ngTemplateOutlet="itemTemplate; context: {
-                                                                $implicit: item
-                                                            }">
-                            </ng-container>
-                        </ng-container>
-                    </ng-container>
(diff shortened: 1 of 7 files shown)


robertandreimoldoveanu (approved):

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
```diff
@@ -187,6 +187,30 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
...
+    @Input()
+    get headerItems() {
+        return this.loading$.value || !!this.inputControl.value.trim()
+            ? []
+            : this._headerItems;
+    }
+    set headerItems(value: ISuggestValue[] | null) {
+        if (!value || isEqual(value, this._headerItems)) { return; }
+
+        if (this.enableCustomValue) {
```
anbalase, line 203: a check should be performed in the setter of enableCustomValue as well

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
```diff
@@ -1190,6 +1250,20 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
...
     };
 
+    private _setAndScrollToActiveIndex(value: ISuggestValue) {
+        const hasSingleSelectValue = !!value && !this.multiple;
+
+        this._setActiveIndex(
+            hasSingleSelectValue
+                ? this.headerItems!.length
+                    ? this.headerItems!.findIndex(({ id }) => id === value.id)
+                    : this.computedItemsOffset(this._items.findIndex(({ id }) => id === value.id))
```
anbalase, line 1260: What if this is called with an item from the `this._items` but also we have `headerItems` defined?
will activeIndex be set correctly?

projects/angular/components/ui-suggest/src/ui-suggest.component.html
```diff
@@ -231,10 +231,59 @@
...
+                                    intl.translateLabel(item.text) }}</span>
+                            </div>
+                            <mat-icon *ngIf="itemExpandable"
+                                      class="expand-icon">chevron_right</mat-icon>
+                        </div>
+                    </ng-template>
+                </mat-list-item>
+            </ng-container>
+
+            <mat-divider *ngIf="headerItems!.length"></mat-divider>
```
anbalase, line 281: can't we avoid template duplication? I'm afraid of maintenance costs when we change anything for items

anbalase: Consider just changing `get renderItems()`.
Divider could be added if `index && index === headerItems!.length`
Am I missing something?

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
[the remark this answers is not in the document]
alexdudescu: nope :D

projects/angular/components/ui-suggest/src/ui-suggest.component.html
alexdudescu: There was indeed an opportunity for extracting the common template section. I did opt as well for altering renderItems in the first place, but that will mess with the rangeLoad directive and accordingly with the requests.

robertandreimoldoveanu (approved):

PR #240, opened 2022-07-27
feat(suggest): add compact summary support

anatolie-darii:
Added a compact mode in order to display the multiselect summary as an alternative to chips
image

Solves: https://uipath.atlassian.net/browse/ACTV-15373

### DIFF
--- a/projects/angular/components/ui-suggest/src/ui-suggest.component.html
+++ b/projects/angular/components/ui-suggest/src/ui-suggest.component.html
@@ -9,7 +9,7 @@
           Tested in NVDA.
       -->
 
-    <ng-container *ngIf="!multiple; else chipsTemplate">
+    <ng-container *ngIf="(!multiple || compact); else chipsTemplate">
         <div [attr.tabindex]="!disabled &&
                         !readonly ?
                         0 : null"
@@ -35,8 +35,7 @@
              matRipple
              class="display"
              role="combobox">
-            <ng-container
-                          *ngIf="itemTemplate && !multiple && value.length && displayTemplateValue; else showDefaultValue">
+            <ng-container *ngIf="itemTemplate && value.length && displayTemplateValue; else showDefaultValue">
                 <ng-container *ngTemplateOutlet="itemTemplate; context: {
                                                 $implicit: value[0]
                                             }">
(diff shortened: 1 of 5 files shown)


projects/angular/components/ui-suggest/src/ui-suggest.component.html
```diff
@@ -9,7 +9,7 @@
           Tested in NVDA.
       -->
 
-    <ng-container *ngIf="!multiple; else chipsTemplate">
+    <ng-container *ngIf="(!multiple || compact); else chipsTemplate">
```
sergiubologa, line 12: Maybe would make sense to extract `(!multiple || compact)` in its own variable in the `*ngLet` above, since it's used 2 times and a suggestive name would make the code easier to read.

sergiubologa (changes requested):
Looks good! Some unit tests and it's good from my side

projects/angular/components/ui-suggest/src/ui-suggest.component.html
anatolie-darii: we should probably do something like refactoring `multiple: boolean` into `multiple: 'chips' | 'compact'` but that would involve more changes and I wanna keep this PR small and safe

sergiubologa (dismissed):

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
```diff
@@ -1105,6 +1119,13 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
         }
     }
 
+    computeDisplayValue() {
+        return this.value.length > 0
+            ? this.value.map(v => this.intl.translateLabel(v.text))
+                .join(',')
```
anbalase, line 1125: could we extract and reuse from [here](https://github.com/UiPath/angular-components/blob/3fb40552de32cd71d6101e4d7aa0208af89b5740/projects/angular/components/ui-suggest/src/ui-suggest.component.ts#L312)? 
```suggestion
                .join(', ')
```

anatolie-darii: done!

anbalase (approved):

sergiubologa (approved):

PR #245, opened 2022-08-08
fix(suggest): do not clear the search input on selecting an item

anatolie-darii:


projects/angular/components/ui-suggest/src/ui-suggest.component.ts
```diff
@@ -1029,12 +1029,11 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
         if (!isItemSelected && value) {
             if (!this.multiple) {
                 this._clearSelection();
-                this._pushEntry(value);
-            } else {
+            } else if (!this.compact) {
                 this.inputControl.setValue('');
```
anatolie-darii, line 1033: @anbalase is there a reason for this `this.inputControl.setValue('');` ? I mean why are we clearing the search input on selection? If the user filters for `Element 1` and gets back the list with `Element 1, Element 11, Element 111` and wants to select maybe 2-3 of them, he'll need to do the filtering 2-3 times since on selecting `Element 1` the search is reset

sergiubologa (approved):

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
anbalase: Yeah, understood, this was a past subject with UX and this is the decision reached.
It seemed like the more likely scenario is to perform a new search after selection rather than persist the current query.
@alexandramehigh do you remember/agree?

alexandramehigh: Correct, I remember the discussion. We've discussed this scenario is connection with for tags > properties. We are using this component for users too. It's more likely to search for a new query instead of continuing to add similar ones.

I imagine the behavior to be similar to the one below.

https://user-images.githubusercontent.com/78076869/183609182-8b72510c-ed11-4b3c-b78c-f6a6c3b77502.mov

anatolie-darii: ok, I didn't change the behaviour in these cases, the only affected case is when you're using the compact mode(as we do in DU)

anbalase (approved):

sergiubologa (approved):

PR #262, opened 2022-09-26
feat(grid): multiple expanded rows

CatalinB7:
adds support for multiple expanded rows

### DIFF
--- a/projects/angular/components/ui-grid/src/ui-grid.component.html
+++ b/projects/angular/components/ui-grid/src/ui-grid.component.html
@@ -240,7 +240,7 @@
                             <ng-container *ngTemplateOutlet="rowTemplate; context: {
                                     data: row,
                                     index: index,
-                                    expanded: expandedEntry,
+                                    expanded: expandedEntries,
                                     last: last
                                 }">
                             </ng-container>
@@ -255,7 +255,7 @@
                         <ng-container *ngTemplateOutlet="rowTemplate; context: {
                                 data: row,
                                 index: index,
-                                expanded: expandedEntry,
+                                expanded: expandedEntries,
                                 last: last
                             }">
                         </ng-container>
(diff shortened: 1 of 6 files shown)


projects/angular/components/ui-grid/src/ui-grid.component.ts
```diff
@@ -318,9 +319,31 @@ export class UiGridComponent<T extends { id: number | string }> extends Resizabl
...
+        return this._expandedEntries[0];
+    }
+
+    /**
+     * Set the expanded entry / entries.
+     *
+     */
+    @Input()
+    set expandedEntries(entry: T | T[]) {
+        if (!entry) {
```
anbalase, line 338: do we want to support this reset mechanism?
if falsy (which isn't declared in the parameter interface) reset expanded?

caleniuc: Yes, this is the current way to reset an expanded row, setting it to `undefined`. I will add this type to the setter parameter.

projects/angular/components/ui-grid/src/ui-grid.component.ts
```diff
@@ -933,6 +963,19 @@ export class UiGridComponent<T extends { id: number | string }> extends Resizabl
...
+        if (rowId == null) {
+            return false;
+        }
+
+        return this._expandedEntries.some(el => el.id === rowId);
+    }
+
+    onRowClick(event: MouseEvent, row: T) {
+        this.rowClick.emit({ event,
+            row });
```
anbalase, line 976:
```suggestion
        this.rowClick.emit({ 
            event,
            row,
             });
```

projects/angular/components/ui-grid/src/ui-grid.component.html
```diff
@@ -280,12 +280,13 @@
...
          class="ui-grid-row"
-         *ngIf="expanded?.id !== row?.id || expandMode === 'preserve'"
-         [class.ui-grid-row-state-expanded]="expanded?.id === row?.id"
+         *ngIf="!isRowExpanded(row?.id) || expandMode === 'preserve'"
+         [class.ui-grid-row-state-expanded]="isRowExpanded(row?.id)"
          [class.ui-grid-border-none]="!footer && last"
          [ngClass]="rowConfig?.ngClassFn(row) ?? ''"
          [tabIndex]="0"
          [attr.data-row-index]="index"
+         (click)="onRowClick($event, row)"
```
anbalase, line 289: we should consider a11y as well, since this is a focusable item, we should probably do the same thing for enter as well.

projects/angular/components/ui-grid/src/ui-grid.component.ts
```diff
@@ -318,9 +319,31 @@ export class UiGridComponent<T extends { id: number | string }> extends Resizabl
...
+        return this._expandedEntries[0];
+    }
+
+    /**
+     * Set the expanded entry / entries.
+     *
+     */
+    @Input()
+    set expandedEntries(entry: T | T[] | undefined) {
+        if (!entry) {
```
anbalase, line 338:
```suggestion
        if (entry != null) {
```

anbalase (approved):

PR #265, opened 2022-10-31
feat(suggest): add a `disabled` flag to ISuggestValue & `displayValueFactory` input

anatolie-darii:
# Changes
1. Add a `disabled` flag to ISuggestValue in order to allow users the possibility to disable items
2. Add the input `displayValueFactory` which will be used both to display the summary and the tooltip for the input. It defaults to `array.join(', ')`


sergiubologa (dismissed):

sergiubologa (approved):

anbalase (approved):
nice!

PR #272, opened 2022-11-01
feat(grid): add support for multiple selection suggest filter

AlexandruCGhimisi:
Add support for multiple selection UiGrid Suggest filter


projects/angular/components/ui-grid/src/ui-grid.component.ts
```diff
@@ -991,6 +993,18 @@ export class UiGridComponent<T extends IGridDataEntry> extends ResizableGrid<T>
         }
     }
 
+    searchableDropdownValue(searchableDropdown: UiGridSearchFilterDirective<T>): ISuggestValue[] {
+        if (searchableDropdown.value) {
+            if (searchableDropdown.multiple) {
+                return searchableDropdown.value! as ISuggestValue[];
+            } else {
+                return [searchableDropdown.value! as ISuggestValue];
```
anbalase, line 1001: doesn't it infer not nullable?
```suggestion
                return searchableDropdown.value as ISuggestValue[];
            } else {
                return [searchableDropdown.value as ISuggestValue];
```

AlexandruCGhimisi: It does indeed. Made the change

anbalase (dismissed):

PR #277, opened 2022-11-09
feat(grid): add label to declare sortable columns

robertandreimoldoveanu:


projects/angular/components/ui-grid/src/ui-grid.intl.ts
```diff
@@ -108,6 +108,10 @@ export class UiGridIntl implements OnDestroy {
      *
      */
     gridHeaderActionsNotice = 'Grid header actions updated. Press Shift + Alt + Arrow Up to move there.';
+    /**
+     * Message to let the users know when they've focused a sortable column
+     */
+     sortableMessage = 'This is a sortable column. Click to sort.';
```
anbalase, line 114: not sure, but will this be breaking for consumers?

robertandreimoldoveanu: hmm indeed it might be. we could instead make it optional and check for existence before displaying to ensure we don't break anything

CatalinB7 (dismissed):

anbalase (approved):

PR #285, opened 2022-11-18
feat(grid): reset search, filters and sorting on action

AlexandruCGhimisi:
* clear search, filters and sorting on action and emit an event to notify user, so they can process the operation

### DIFF
--- a/projects/angular/components/ui-grid/src/ui-grid.component.ts
+++ b/projects/angular/components/ui-grid/src/ui-grid.component.ts
@@ -781,6 +781,19 @@ export class UiGridComponent<T extends IGridDataEntry> extends ResizableGrid<T> 
     }
 
     /**
+     * Clear search term, filters and sorting and emits true after.
+     */
+    @Input()
+    reset: () => Observable<boolean> = () => {
+        if (this.header) {
+            this.header.searchValue = '';
+        }
+        this.filterManager.clear();
+        this.sortManager.clear();
+        return of(true);
+    };
+
+    /**
      * @ignore
      */
     ngAfterContentInit() {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# v13.13.0 (2022-11-23)
+* **grid** add overridable input that reset grid state and emits after (by default it clears search term, filters and sorting)
+
 # v13.12.5 (2022-11-22)
 * **dateformat** avoid using macrotasks using setTimeout generated a lot of macrotasks
 * **test** update test naming
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "angular-components",
-  "version": "13.12.5",
+  "version": "13.13.0",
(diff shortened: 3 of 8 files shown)


projects/playground/src/app/pages/grid/component/grid.component.html
```diff
@@ -37,6 +37,14 @@
             </ng-template>
         </ui-header-button>
 
+        <ui-header-button type="inline">
```
CalinaCristian, line 40: this should be an input, not always display it

projects/playground/src/app/pages/grid/component/grid.component.ts
```diff
@@ -85,7 +85,14 @@ export class GridComponent implements OnInit, OnDestroy, AfterViewInit {
         this.paginateData(this.filteredData, this.pageIndex, this.footer.pageSize);
     }
 
+    reset() {
+        this._grid.resetAll();
+    }
+
     ngAfterViewInit() {
+        this._grid.refresh.subscribe(_ =>
```
CalinaCristian, line 93: this should be under a `takeUntil(this._destroyed$)`.. Also, why adding this logic here? Shouldn't it be done in `reset()` method above?

CalinaCristian (commented):
You should add tests for this scenario as well.

projects/angular/components/ui-grid/src/ui-grid.component.ts
```diff
@@ -965,6 +972,15 @@ export class UiGridComponent<T extends IGridDataEntry> extends ResizableGrid<T>
         this.filterManager.clearCustomFilters();
     }
 
+    clearAll() {
```
anbalase, line 975: would be nice if consumers can override this behavior, it should also return a stream so you can compose reactive logic to it.
e.g. 
```ts
// [clearAll]=() { return this._consumerService.resetStorage(); }

@Input()
clearAll: (this)=> Observable<boolean>  = ...

// ...
clearButtonClick() {
    this.clearAll().pipe(
    tap(()=>this.filtersCleared.emit()),
).subscribe();
}

```

anbalase, line 975: how should a consumer hook this up? 
should we expose a standard inline clear button that invokes this?
Y - pls add
N - let's add a comment with for recommended usage.

anbalase (changes requested):

projects/angular/components/ui-grid/src/ui-grid.component.ts
AlexandruCGhimisi: On the long run it would be nice to have a standard button. This requires design validation, so I would skip that for when we really need this functionality. For now, we only need a way to programmatically clear filters.

AlexandruCGhimisi: Nice idea. I'll do that

anbalase (approved):

anbalase (approved):

PR #287, opened 2022-11-22
fix(dateformat): call markForCheck only on initial assignment

anbalase:
markForCheck is needed so that changes propagate to matTooltip


projects/angular/directives/ui-dateformat/src/ui-dateformat.directive.spec.ts
```diff
@@ -531,15 +531,17 @@ describe('Directive: UiDateFormat', () => {
             expect(momentOutputDate.minute()).toEqual(momentInputDate.minute());
         });
 
-        it('should call detectChanges if the date input value changes', () => {
+        it('should call detectChanges if the date input value changes', fakeAsync(() => {
```
iordancatalin, line 534: nit:

```suggestion
        it('should call markForCheck if the date input value changes', fakeAsync(() => {
```

iordancatalin (approved):

iordancatalin (approved):

PR #289, opened 2022-11-23
feat/tree select

robertandreimoldoveanu:


projects/angular/components/ui-tree-select/src/ui-tree-item/ui-tree-item.component.html
```diff
@@ -0,0 +1,33 @@
+<!-- <mat-list-item class="node">
+
+  <span>
+    {{ node.name }}
+  </span>
+
+</mat-list-item>
+
+ -->
```
alexdudescu, line 9: 💬

projects/angular/components/ui-tree-select/src/ui-tree-select.component.scss
```diff
@@ -0,0 +1,48 @@
+.virtual-scroll-container {
+  border: 2px solid cyan;
+  height: 500px;
+  overflow: auto;
+}
+
+// .node {
+//   display: block;
+//   height: 24px;
+// }
```
alexdudescu, line 10: 💬

projects/angular/components/ui-tree-select/src/ui-tree-select.component.scss
```diff
@@ -0,0 +1,48 @@
...
+  &.selected {
+    background-color: gold;
+  }
+}
+
+@mixin perLevelPadding($level, $padding) {
+  .node-level-#{$level} {
+
+    &:not(.node-type-leaf) {
+      padding-left: max(0, calc($padding * ($level)));
```
alexdudescu, line 34: Can you ever have negative levels? As for the padding I think you there's only negative margin.
```suggestion
      padding-left: calc($padding * ($level));
```

projects/playground/src/app/pages/tree-select/tree-select.page.ts
```diff
@@ -0,0 +1,117 @@
...
+//       children.push(generateNode(level + 1, i));
+//     }
+//   }
+//   const name = Array(level + 1).fill(0).map(() => String.fromCharCode(65 + index)).join('');
+//   return {
+//     key: (idx++).toString(),
+//     name: `Folder ${name}`,
+//     children,
+//   };
+// }
```
alexdudescu, line 66: 💬

projects/angular/components/ui-tree-select/src/models/tree.ts
```diff
@@ -0,0 +1,14 @@
+export interface Node {
+    key: string;
+    name: string;
+    children?: Node[];
+    data?: any;
+}
+
+export interface FlatNode {
```
anbalase, line 8: would love to have this interface generic for data, by default set to any.

projects/angular/components/ui-tree-select/src/ui-tree-item/ui-tree-item.component.html
```diff
@@ -0,0 +1,33 @@
...
+     [ngClass]="'node-level-' + node.level"
+     [class.node-type-leaf]="!node.hasChildren"
+     class="tree-item-container"
+     matRipple
+     [class.selected]="isSelected">
+    <button *ngIf="node.hasChildren"
+            uiSpinnerButton
+            mat-button
+            [spinnerButtonLoading]="loading"
+            class="toggle-button"
```
anbalase, line 21: `type="button"`

projects/angular/components/ui-tree-select/src/ui-tree-item/ui-tree-item.component.html
```diff
@@ -0,0 +1,33 @@
...
+            tabindex="-1">
+        <mat-icon>{{ !isExpanded ? 'expand_more' : 'chevron_right' }}</mat-icon>
+    </button>
+
+    <mat-list-item class="node">
+
+        <span>
```
anbalase, line 28: is this needed?

projects/angular/components/ui-tree-select/src/ui-tree-select.component.html
```diff
@@ -0,0 +1,21 @@
+<ng-container *ngIf="loadingFolderExpansion$ | async as loadingFolder">
+    <cdk-virtual-scroll-viewport (keydown)="onKeydown($event)"
+                                 class="virtual-scroll-container"
+                                 tabindex="0"
+                                 itemSize="26">
+        <mat-list class="tree-container">
+            <ng-container *ngIf="dataSource.data.length == 0">
+                <ng-content select="[noDataMessage]"></ng-content>
```
anbalase, line 8: 👍

projects/angular/components/ui-tree-select/src/ui-tree-select.component.scss
```diff
@@ -0,0 +1,48 @@
+.virtual-scroll-container {
+  border: 2px solid cyan;
```
anbalase, line 2: ups

projects/angular/components/ui-tree-select/src/ui-tree-select.component.scss
```diff
@@ -0,0 +1,48 @@
...
+
+    &.node-type-leaf {
+      padding-left: 32px + max(32px, calc($padding * ($level + 1)));
+    }
+  }
+}
+
+
+.tree-container {
+  @include perLevelPadding(0, 32px);
```
anbalase, line 45: https://sass-lang.com/documentation/at-rules/control/for

projects/angular/components/ui-tree-select/src/models/tree.ts
```diff
@@ -0,0 +1,14 @@
+export interface Node {
```
anbalase, line 1: consider a more specific naming for interfaces we don't want to confuse these

projects/angular/components/ui-tree-select/src/models/tree.ts
```diff
@@ -0,0 +1,14 @@
...
+    key: string;
```
anbalase, line 2: key should support both `string | number`

projects/angular/components/ui-tree-select/src/ui-tree-select.component.ts
```diff
@@ -0,0 +1,157 @@
...
+    get dataSource() {
+        return this._dataSource;
+    }
+
+    loading$ = new BehaviorSubject(true);
+    loadingFolderExpansion$ = new BehaviorSubject<LoadingFolderExpansion>({ value: false });
+    expandedNodes: FlatNode[] = [];
+    currentSelectedNode: FlatNode | null = null;
+
+    private keyManager?: FocusKeyManager<UiTreeItemComponent>;
```
anbalase, line 76:
```suggestion
    private _keyManager?: FocusKeyManager<UiTreeItemComponent>;
```

anbalase (changes requested):
pls add decent coverage

projects/angular/components/ui-tree-select/src/ui-tree-select.component.scss
robertandreimoldoveanu: fair point.
took note of the feedback and moved the styles to the playground.

projects/angular/components/ui-tree-select/src/ui-tree-item/ui-tree-item.component.html
robertandreimoldoveanu: nope it wasn't, removed it.
thanks

projects/angular/components/ui-tree-select/src/ui-tree-select.component.scss
robertandreimoldoveanu: thanks for the tip, was thinking that we might need a for loop there 😄 
moved the styles to the playground tho, as I feel these are consumer decisions to be made.
could we export this mixin to the consumer? so they can just call it with their own padding values if needed

projects/angular/components/ui-tree-select/src/models/tree.ts
robertandreimoldoveanu: changed to `IFlatNodeObject` 'cause it's an interface the consumer shouldn't need, and `ITreeNode` which is self explanatory and seems better formulated

projects/angular/components/ui-tree-select/src/models/tree.ts
robertandreimoldoveanu: do we want that though? consumers with IDs could `toString` and have certainty of type
thinking that we've had situations where we wanted to support both and were problematic 😓

anbalase: Well, you're just pushing this issue to consumers. 
But, since this won't be used as a CVA and output doesn't matter, I guess it's fine to keep these as strings here.

projects/angular/components/ui-tree-select/src/models/tree.models.ts
```diff
@@ -0,0 +1,14 @@
...
+    data?: T;
+}
+
+export interface IFlatNodeObject<T = any> {
+    key: string;
+    name: string;
+    hasChildren: boolean;
+    level: number;
+    data?: T;
+}
```
CalinaCristian, line 14: can't we extract the common `key`, `name` and `data` to an interface and extend it in both of those interfaces?

anbalase (changes requested):

projects/angular/components/ui-tree-select/src/ui-tree-item/ui-tree-item.component.html
```diff
@@ -0,0 +1,22 @@
...
+          (click)="toggle()"
+          uiSpinnerButton
+          mat-button
+          class="toggle-button"
+          type="button"
+          tabindex="-1">
+    <mat-icon>{{ !isExpanded ? 'expand_more' : 'chevron_right' }}</mat-icon>
+  </button>
+
+  <mat-list-item (click)="click()"
```
anbalase, line 18: pls check a11y wise how this behaves without a mat-list as a direct parent

projects/angular/components/ui-tree-select/src/ui-tree-select.component.html
```diff
@@ -0,0 +1,22 @@
+<ng-container *ngIf="loadingFolderExpansion$ | async as loadingFolder">
+    <cdk-virtual-scroll-viewport (keydown)="onKeydown($event)"
+                                 class="virtual-scroll-container"
+                                 tabindex="0"
+                                 itemSize="26">
+        <mat-list class="tree-container">
```
anbalase, line 6:
```suggestion
        <mat-list class="tree-container" role="list">
```

projects/angular/components/ui-tree-select/src/ui-tree-select.component.html
```diff
@@ -0,0 +1,22 @@
+<ng-container *ngIf="loadingFolderExpansion$ | async as loadingFolder">
+    <cdk-virtual-scroll-viewport (keydown)="onKeydown($event)"
+                                 class="virtual-scroll-container"
+                                 tabindex="0"
+                                 itemSize="26">
```
anbalase, line 5: should this be overridable based on the template used?

projects/angular/components/ui-tree-select/src/ui-tree-select.component.ts
```diff
@@ -0,0 +1,159 @@
...
+})
+export class UiTreeSelectComponent implements AfterViewInit {
+
+    @ViewChildren(UiTreeItemComponent)
+    items!: QueryList<UiTreeItemComponent>;
+
+    @Input()
+    itemSize = 26;
+
+    @HostBinding('style.--ui-tree-select-item-padding')
```
anbalase, line 41: 🤟

CalinaCristian: 😮

projects/angular/components/ui-tree-select/src/ui-tree-select.component.html
```diff
@@ -0,0 +1,31 @@
...
+            <ng-container *ngIf="dataSource.data.length == 0">
+                <ng-content select="[noDataMessage]"></ng-content>
+            </ng-container>
+            <ng-container *cdkVirtualFor="let node of dataSource; let i = index; trackBy: trackById;">
+                <ui-tree-item (selected)="select(node, i)"
+                              (collapsed)="collapse(node)"
+                              (expanded)="expand(node)"
+                              [node]="node"
+                              [isSelected]="isSelected(node)">
+                    <ng-template *ngIf="itemTemplate"
```
anbalase, line 17: It's very powerful to just allow the consumer to specify a template. 
Shouldn't we have a default, still? 
Seems a bit of an overhead that each consumer needs to implement this if you just need to render the text of the node.

projects/angular/components/ui-tree-select/src/ui-tree-select.component.ts
```diff
@@ -0,0 +1,193 @@
...
+            value: false,
+        });
+    }
+
+    get dataSource() {
+        return this._dataSource;
+    }
+
+    loadingFolderExpansion$ = new BehaviorSubject<LoadingFolderExpansion>({ value: false });
+    currentSelectedNode: IFlatNodeObject | null = null;
```
anbalase, line 102: how will this work for multi-select?
are going to rewrite it for that scenario?

anbalase: why not have a set of selected nodes?

anbalase: do we care to hydrate the selection from the consumer?

package.json
```diff
@@ -34,6 +34,7 @@
     "build:theme": "scss-bundle angular -c scss-bundle.config.json",
     "build:readme": "cpx ./README.md ./dist/angular",
     "build:prod": "npm-run-all lint build build:theme build:readme",
+    "test": "npm run test:watch",
     "test:watch": "ng test",
```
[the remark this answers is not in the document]
robertandreimoldoveanu: no
`test` is just a shortcut to the existing `test:watch` command. the `watch` is default
now it can be ran with `npm test` instead of `npm run test:watch`

projects/angular/components/ui-tree-select/src/ui-tree-select.component.html
robertandreimoldoveanu: I think that supporting both right now adds some convoluted logic in the template. i'd rather we have simplicity than turn into another `ui-suggest`. 

the most basic use would be:
```
<ng-template *ngIf="itemTemplate" let-node="node">
     {{node.name}}
</ng-template>
```

we can hardly call that overhead. on the other hand, implementing a default and asking the consumer to provide an `Intl` service is more overhead imho than this

projects/angular/components/ui-tree-select/src/ui-tree-select.component.ts
robertandreimoldoveanu: i think multiselect would be a v2 feature. surely needs to exist, but not right now
i missed hydrating the selection, will add

projects/angular/components/ui-tree-select/src/ui-tree-select.intl.ts
```diff
@@ -0,0 +1,36 @@
...
+} from '@angular/core';
+import { Subject } from 'rxjs';
+
+/**
+ * Internationalization service definition for `UiTreeSelect`.
+ *
+ * @export
+ */
+@Injectable()
+export class UiTreeSelectIntl implements OnDestroy {
```
anbalase, line 12: drop

projects/angular/components/ui-tree-select/src/ui-tree-select.component.ts
```diff
@@ -0,0 +1,217 @@
...
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    encapsulation: ViewEncapsulation.None,
+})
+export class UiTreeSelectComponent implements AfterViewInit {
+
+    @ViewChildren(UiTreeItemComponent)
+    items!: QueryList<UiTreeItemComponent>;
+
+    @ContentChild('itemTemplate')
+    itemTemplate!: any;
```
anbalase, line 59:
```suggestion
    itemTemplate?: TemplateRef<any>;
```

anbalase (changes requested):

robertandreimoldoveanu:
the `check-public-exports` script failed due to:

image

the job running this script gets skipped under some conditions. I expect during the Angular 14 upgrade it became deprecated but still somehow got skipped

projects/playground/src/app/pages/tree-select/tree-select.module.ts
```diff
@@ -0,0 +1,24 @@
...
+  imports: [
+    CommonModule,
+    UiTreeSelectComponent,
+    MatFormFieldModule,
+    MatIconModule,
+    UiSpinnerButtonModule,
+    MatButtonModule,
+  ],
+})
+export class TreeSelectModue { }
```
anbalase, line 24: do we need this?

projects/angular/components/ui-tree-select/src/ui-tree-select.component.ts
```diff
@@ -0,0 +1,234 @@
...
+
+    @Input()
+    itemSize = 26;
+
+    /**
+     * Path of keys key to the node that will be selected by default. The node must be present in the data array
+     * If the node is a lower level node, the parent nodes must be present in the data array and then they will be expanded automatically
+     */
+    @Input()
+    set initialSelection(value: string[]) {
```
anbalase, line 68:
```suggestion
    set initialSelection(value: string[] | null) {
```

anbalase (approved):

anbalase (approved):

PR #296, opened 2023-01-10
fix/improve tree select

robertandreimoldoveanu:


projects/angular/components/ui-tree-select/src/ui-tree-select.component.html
```diff
@@ -8,7 +8,7 @@
             <ng-content select="[noDataMessage]"></ng-content>
         </ng-container>
         <ng-container *cdkVirtualFor="let node of dataSource; let i = index; trackBy: trackById;">
-            <ui-tree-item (selected)="select(node, i)"
+            <ui-tree-item (selected)="select(node)"
```
iordancatalin, line 11: nit: remove `i` from `*ngFor` if no longer used

projects/angular/components/ui-tree-select/src/ui-tree-select.component.ts
```diff
@@ -119,14 +130,53 @@ export class UiTreeSelectComponent implements AfterViewInit {
...
     private _initialSelection: string[] = [];
 
+    private _connectedData$ = new BehaviorSubject<IFlatNodeObject[]>([]);
+    private _selectNode$ = new Subject<IFlatNodeObject>();
+    private _destroyed$ = new Subject<void>();
+
+    constructor(private _cd: ChangeDetectorRef) {
+        this.dataSource.connect({ viewChange: of(null) as any }).subscribe(a => {
+            this._connectedData$.next(a);
+        });
```
iordancatalin, line 140: Can `of(null) as any` be replaced with `EMPTY`?
Don't you need to unsubscribe from this stream?
Change the name of the parameter from subscribe, `a` is too generic.

iordancatalin (dismissed):

iordancatalin (approved):

PR #301, opened 2023-01-26
feat(grid): add minchars and displaycount to search filter

CalinaCristian:


anbalase (changes requested):
code looks good, pls add relevant tests

projects/angular/components/ui-grid/src/models/filterModel.ts
```diff
@@ -28,4 +30,9 @@ export interface IFilterModel<T> {
      *
      */
     type?: string;
+    /**
+     * The current filter meta.
```
anbalase, line 34: could you pls explain when this should be specified and why? what's the usecase?

CalinaCristian: this is additional metadata that filter change will send. It doesn't need to be specified, it's just extra data in case you need more than the id.

CalinaCristian: I added more information on the comment.

anbalase (approved):

projects/angular/components/ui-grid/src/models/filterModel.ts
anbalase: thanks

PR #302, opened 2023-01-26
feat: add opt-in to use Luxon instead of Moment [OR-62547]

seferturan:
## 🛠 Fixes 🛠
Partial fix for https://uipath.atlassian.net/browse/OR-62547; 2 more repos to go.

Where applicable, adds an opt-in to use Luxon instead of Momentjs.

## 🎶 Notes 🎶
While similar, there are some differences:

- `humanize-duration` is used to make `Duration` more readable. This is not 100% the same as momentjs; e.g. where `momentjs` formats it as `A second ago`, it will now be `1 second ago`.
- For any formatting intended to be read by a human, `Luxon` recommends using `toLocaleString`. The following presets should be usable: https://github.com/moment/luxon/blob/master/docs/formatting.md#presets
- ~Right now, every component has its own token. We could also create 1 token `angular-internal` for all cases~


caleniuc:
Please also bump the version after the tests are fixed, here are the steps:

1. `npm run bump:minor` (other version options for the command: `patch`, `major`, `stable`, `rc`, `patch:rc`, `minor:rc`, `major:rc`)
2. `git push` 
3. `git push origin <tagname>`
4. After the PR is merged, switch to master branch
4.1 `npm ci`
4.2. `npm run build:prod`
4.3. `npm run publish:gh`
4.4. `npm run publish:npm`
5. [draft a new release](https://github.com/UiPath/angular-components/releases), use changelog message as description, choose tag

projects/angular/utilities/src/luxon/useLuxon.ts
```diff
@@ -0,0 +1,8 @@
+import { InjectionToken } from '@angular/core';
```
anbalase, line 1: nit: we prefer kebab case naming for files though not an enforced rule

caleniuc (dismissed):

anbalase (approved):

PR #304, opened 2023-01-30
feat(suggest): add icons on chips

CalinaCristian:
Adds support for icons in chips.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# v14.2.0 (2023-01-30)
+* **grid** add icon on chips
+* **fix** version bump script
+* **chore** bump version to 14.1.0
+
 # v14.1.0 (2023-01-27)
 * **test** adds tests for luxon compatible directives
 * **feat** use luxon as an opt-in feature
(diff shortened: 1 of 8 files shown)


projects/angular/components/ui-suggest/src/ui-suggest.component.html
```diff
@@ -102,6 +102,14 @@
                           [matTooltip]="intl.translateLabel(option.text)"
                           (removed)="deselectItem(option)"
                           class="chip-selectable">
+                    <mat-icon *ngIf="!!option?.icon?.svgIcon && displayIconInChip"
```
anbalase, line 105: If consumers already provide `option.icon` why should we also add a flag for rendering?
wdyt?

anbalase: My first thought would be to not flag this added feature.

anbalase: maybe switch the flag to true by default to keep the logic but have it implicit as a recommendation. This is a cosmetic bonus and not a breaking change.

CalinaCristian: I'll remove the flag entirely 👍

anbalase (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# v14.2.0 (2023-01-30)
+* **grid** add icon on chips
```
anbalase, line 2: nit: it's actually a suggest feat right :D

PR #305, opened 2023-01-31
Fix(keyboard-shortcut): require keys to be pressed again in order to emit again

CatalinB7:
when all the required keys for a shortcut are pressed simultaneously an event is emitted and the key pressed state is reset


projects/angular/directives/keyboard-shortcut/src/keyboard-shortcut.directive.ts
```diff
@@ -32,6 +32,7 @@ export class KeyboardShortcutDirective {
         this._pressedKeys[event.key] = true;
         if (this.shortcutKeys.find(keyCombination => keyCombination.every(key => this._pressedKeys[key]))) {
             this.shortcutPressed.emit();
+            Object.keys(this._pressedKeys).forEach(k => this._pressedKeys[k] = false);
```
anbalase, line 35: teeest?

anbalase: why not `this._pressedKeys={}`

anbalase (approved):

PR #309, opened 2023-02-07
Feat/custom value template

alexdudescu:
### Feature description

Adds `applyItemTemplateToCustomValue` input which allows the user to control to switch between the default custom value template or sharing the provided `itemTemplate`

### Usage

- set `applyItemTemplateToCustomValue`
- in template you have access to `isCustomValue` to determine whether the current item is existing or a custom value
- you also have access to `isCustomValueAlreadySelected` to determine whether the current custom value is currently selected or not
- note that by default the item passed via context for an existing item will be the item's data, while for a custom value it will be the value of the input 

```javascript
<ui-suggest [enableCustomValue]="true"
            [applyItemTemplateToCustomValue]="true"
            ... >
     <ng-template let-item
                  let-isCustomValue="isCustomValue"
                  let-isCustomValueAlreadySelected="isCustomValueAlreadySelected">

        <ng-container *ngIf="!isCustomValue">
            {{item.text}}
        </ng-container>

        <ng-container *ngIf="isCustomValue && !isCustomValueAlreadySelected">
            + Add {{item}}
        </ng-container>

        <ng-container *ngIf="isCustomValue && isCustomValueAlreadySelected">
            Value already selected
        </ng-container>
    </ng-template>
</ui-suggest>
```

### DIFF
--- a/projects/angular/components/ui-suggest/src/ui-suggest.component.ts
+++ b/projects/angular/components/ui-suggest/src/ui-suggest.component.ts
@@ -189,6 +189,13 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
     ignoreOpenOnFetch = false;
 
     /**
+     * Controls whether to use the default custom value template or use the custom item template
+     *
+     */
+    @Input()
+    applyItemTemplateToCustomValue = false;
+
+    /**
      * A list of options that will be presented at the top of the list.
      *
      */
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+# v14.3.0 (2023-02-08)
+* **playground** showcase custom value template support
+* **suggest** add custom value template support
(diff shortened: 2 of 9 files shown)


projects/angular/components/ui-suggest/src/ui-suggest.component.ts
```diff
@@ -188,6 +188,13 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
     @Input()
     ignoreOpenOnFetch = false;
 
+    /**
+     * Controls weather to use the default custom value template or use the custom item template
+     *
+     */
+    @Input()
+    useCutomValueTemplate = false;
```
anbalase, line 196:
```suggestion
    applyItemTemplateToCustomValue = false;
```

anbalase (dismissed):

anbalase (approved):

PR #312, opened 2023-03-01
Fix: display concatenated tooltip (title + description) on keyboard focus

CatalinB7:
When the column header is focused it should display a tooltip that consists of the concatenation of column title and column description.
Demo: 

https://user-images.githubusercontent.com/87014385/222734741-12d3ef1b-e2c2-4f30-9337-0d36b061baab.mp4

### DIFF
--- a/projects/angular/components/ui-grid/src/ui-grid.component.html
+++ b/projects/angular/components/ui-grid/src/ui-grid.component.html
@@ -139,7 +139,6 @@
 
                     <ng-container *ngIf="multiPageSelect then multiPageSelectionProjectedHeaderCell">
                     </ng-container>
-
                     <div *ngFor="let column of visible$ | async; let last = last; let columnIndex = index"
                          [class.ui-grid-header-cell-sortable]="column.sortable"
                          [class.ui-grid-resizeable]="column.resizeable"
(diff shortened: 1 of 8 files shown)


CalinaCristian (approved):
have you verified that it still looks ok with both sortable and description (are they aligned properly now that the DOM structure changed)?

anbalase (changes requested):
a tooltip would be needed on focus as well

caleniuc:
On blur on a column with the info icon displayed, right after the big tooltip is hidden, it seems it shows also the `p` tag's tooltip only for a very short period.

projects/angular/components/ui-grid/src/ui-grid.component.html
```diff
@@ -152,23 +152,29 @@
                          (keyup.space)="sortManager.changeSort(column)"
                          (keydown.ArrowLeft)="!last && resizeManager.startKeyboardResize('left', column, columnIndex)"
                          (keydown.ArrowRight)="!last && resizeManager.startKeyboardResize('right', column, columnIndex)"
+                         (focus)="focusedColumnHeader = true; columnTooltip.show()"
+                         (focusout)="focusedColumnHeader = false; columnTooltip.hide()"
```
CatalinB7, line 156:
```suggestion
                         (focusout)="focusedColumnHeader = false; columnTooltip.hide(0)"
```

CatalinB7: keyboard focus only

projects/angular/components/ui-grid/src/ui-grid.component.ts
```diff
@@ -1043,6 +1051,16 @@ export class UiGridComponent<T extends IGridDataEntry> extends ResizableGrid<T>
...
     }
 
+    triggerColumnHeaderTooltip(event: FocusEvent, tooltip: MatTooltip) {
+        setTimeout(() => {
+            const styles = window.getComputedStyle(event.target as HTMLElement, ':before');
+            if (styles.backgroundColor === KEYBOARD_FOCUSED_ELEMENT_BACKGROUND_COLOR) {
+                this.focusedColumnHeader = true;
+                tooltip.show();
+            }
+        }, 0);
```
CatalinB7, line 1061: focus-visible appears for a short moment on mouse clicks as well :(

robertandreimoldoveanu (changes requested):
the solutions is ingenious but convoluted and brings no value as all a11y criteria are met w/o it. i can't see any valuable use for this other than a tiny nice to have. imho it's code bloat. 

i would've proposed adding some tests but that would make it even worse.

can we revisit this?

projects/angular/components/ui-grid/src/ui-grid.component.ts
```diff
@@ -1043,6 +1051,21 @@ export class UiGridComponent<T extends IGridDataEntry> extends ResizableGrid<T>
             || (column.searchableDropdown?.value != null && (column.searchableDropdown?.value as ISuggestValue[])?.length !== 0);
     }
 
+    triggerColumnHeaderTooltip(event: FocusEvent, tooltip: MatTooltip) {
+        setTimeout(() => {
+            const styles = window.getComputedStyle(event.target as HTMLElement, ':before');
```
CalinaCristian, line 1056: I agree with @robertandreimoldoveanu. It seems like the voiceover utility reads the whole tooltip when you focus via keyboard (both description and title). You don't need to show the tooltip in this scenario.

Also, this `getComputedStyle` method is known to be slow.
You can read more about it [here](https://gist.github.com/paulirish/5d52fb081b3570c81e3a#calling-getcomputedstyle)

> Calling getComputedStyle()
window.getComputedStyle() will typically force style recalc.
window.getComputedStyle() will often force layout, as well.

CalinaCristian: Screenshot 2023-03-02 at 4 11 54 PM

CatalinB7:
it is a requirement for the information to be available to someone who only uses the keyboard => a tooltip must be displayed to show the column description => 2 possibilities:
1. make the icon focusable 
2. display tooltip with the concatenation of column title and description.

Will try to implement the 2nd.

robertandreimoldoveanu (approved):
still unsure about the worth of this but at least the code looks manageable and is cleaner

anbalase (commented):
@CatalinB7 pls specify the a11y criteria for clarity. 
How does this logic behave for the screen-reader when focusing a column header?

CatalinB7:
https://sarahmhigley.com/writing/tooltips-in-wcag-21/
"Tooltips must be discoverable and readable with a mouse, other pointer devices, keyboard, screen reader, zoom software, and any other assistive technology."

> How does this logic behave for the screen-reader when focusing a column header?

https://user-images.githubusercontent.com/87014385/223557586-e2c65f54-40c6-4f19-8d00-a8fd9c198f8a.mp4

anbalase (approved):

alexdudescu (approved):

PR #315, opened 2023-03-10
feat(suggest): adding cdk layout for dropdown

ZoicanDenisUi:
**Description**
If we place the `ui-suggest` component inside a container that has `overflow:hidden`, the user cannot see the options and cannot interact with them. To solve this problem, I used `cdk layout` to show the options that will be placed over all elements.

I added a new Input property that will forcefully place the dropdown over the component.

**Breaking change**
A problem that users may encounter is that the scss that was applied on the elements is not visible anymore because the dropdown was moved outside the component and class names were changed.  To solve this problem, the scss queries must be changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# v14.6.0 (2023-03-16)
+* **suggest** use CDK Layout to display the dropdown (breaking change)
+
 # v14.5.4 (2023-03-08)
 * **grid** add aria-expanded to filters btn
 * **grid** add tick to test
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "angular-components",
-  "version": "14.5.4",
+  "version": "14.6.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "angular-components",
-      "version": "14.5.4",
+      "version": "14.6.0",
       "license": "MIT",
       "dependencies": {
         "@angular/animations": "14.2.12",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "angular-components",
-  "version": "14.5.4",
+  "version": "14.6.0",
   "author": {
     "name": "UiPath Inc",
     "url": "https://uipath.com"
--- a/projects/angular/package.json
+++ b/projects/angular/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@uipath/angular",
-    "version": "14.5.4",
+    "version": "14.6.0",
     "license": "MIT",
     "author": {
         "name": "UiPath Inc",


projects/angular/components/ui-suggest/src/ui-suggest.component.html
```diff
@@ -138,106 +140,113 @@
...
-            <ng-container *ngIf="!isDown">
-                <ng-container *ngTemplateOutlet="searchableCountInfoTmpl"></ng-container>
+</div>
 
-                <ng-container *ngTemplateOutlet="resultList">
-                </ng-container>
+<ng-template #optionsDropdown>
+    <div [style.width]="suggestContainerWidth+'px'"
+         [style.height]="expandInline ? '100%' : undefined"
+         [class.suggest-dropdown-item-list-container-expanded-inline]="expandInline"
```
sergiubologa, line 161:
```suggestion
         [class.ui-suggest-dropdown-item-list-container-expanded-inline]="expandInline"
```

projects/angular/components/ui-suggest/src/ui-suggest.component.spec.ts
```diff
@@ -2154,6 +2165,8 @@ const sharedSpecifications = (
             it(`should fetch call after the 'minChars' is met`, waitForAsync(async () => {
                 const MIN_CHARS = 5;
                 component.minChars = MIN_CHARS;
+                component.searchable = true;
+                component.multiple = true;
```
sergiubologa, line 2169: why are these needed?

projects/angular/components/ui-suggest/src/ui-suggest.component.spec.ts
```diff
@@ -418,6 +418,7 @@ const sharedSpecifications = (
             const items = generateSuggetionItemList(25);
 
             component.searchable = true;
+            component.multiple = true;
```
sergiubologa, line 421: is this really needed?

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
```diff
@@ -765,6 +785,18 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
...
 
+        if (this.isFormControl) {
+            this.upPosition = {
+                ...this.upPosition,
+                offsetY: -10,
+            };
+            this.downPosition = {
+                ...this.downPosition,
+                offsetY: 10,
+            };
```
sergiubologa, line 796: please move this into a private function with a descriptive name (e.g.: `_adjustDropdownPositionWhenUsingFormControl()`). Also, please add a comment explaining why this adjustment is needed.

projects/angular/components/ui-suggest/src/ui-suggest.component.html
```diff
@@ -138,106 +140,113 @@
...
-             tabindex="-1"
-             class="item-list-container">
-            <ng-container *ngIf="!isDown">
-                <ng-container *ngTemplateOutlet="searchableCountInfoTmpl"></ng-container>
+</div>
 
-                <ng-container *ngTemplateOutlet="resultList">
-                </ng-container>
+<ng-template #optionsDropdown>
+    <div [style.width]="suggestContainerWidth+'px'"
```
sergiubologa, line 159:
```suggestion
    <div [style.width.px]="suggestContainerWidth"
```

sergiubologa (changes requested):

sergiubologa (changes requested):
There's also a small difference between the master and your branch, in the way the dropdown/overlay is displayed when opened, in relation to the parent control. For example, for the first item in the playground (Single select suggest) this is how it's opened on the master branch
image
and this is how it's opened on your branch:
image

Notice that on the master the overlay opens up over the parent control, covering it. On your branch, it opens up under the parent control.

projects/angular/components/ui-suggest/src/ui-suggest.component.html
ZoicanDenisUi: Changed it.

projects/angular/components/ui-suggest/src/ui-suggest.component.spec.ts
ZoicanDenisUi: In the old implementation, the test would see the input from the dropdown that would have a height equal to zero. To solve this problem, by setting this property, an input would be rendered in the component and the test will check what it needs.

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
ZoicanDenisUi: Refactored the code and moved this logic inside a function.

projects/angular/components/ui-suggest/src/ui-suggest.component.html
ZoicanDenisUi: Changed it

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
```diff
@@ -845,6 +879,10 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
     }
 
     ngOnChanges(changes: SimpleChanges) {
+        if (changes.direction) {
+            this.dropdownPosition = [this.direction === 'up' ? this.upPosition : this.downPosition];
```
sergiubologa, line 883: you should call here the `_initDropdownPositions` function

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
```diff
@@ -229,6 +229,12 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
     @Input()
     expandInline = false;
 
+    /**
+     * If true, component wil place the dropdown over the input
+     */
+    @Input()
+    forceOriginDropdownPosition = false;
```
sergiubologa, line 236:
```suggestion
    forceDisplayDropdownOverInput = false;
```

projects/angular/components/ui-suggest/src/ui-suggest.animations.ts
```diff
@@ -19,18 +19,23 @@ export const UI_SUGGEST_ANIMATIONS: {
...
+            'void => open',
+            group([
+                animate('100ms linear', style({
+                    opacity: 1,
+                })),
+                animate('120ms cubic-bezier(0, 0, 0.2, 1)', style({
+                    transform: 'scale(1)',
+                })),
+            ]),
+        ),
```
sergiubologa, line 39: 👍 I agree, it looks better without the close transition.

sergiubologa (changes requested):
When displayed above and over the parent it's not properly aligned:
image

sergiubologa:
There's another use case that needs to be considered: `ui-suggest` is also used in the `ui-grid`, for filtering. This is how it looks now:
image
this is how it looks on master:
image

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
```diff
@@ -1242,6 +1282,40 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
...
+                ...this.upPosition,
+                offsetX: -1,
+                offsetY: mustBePlacedOutsideTheInput ? -10 : 11,
+            };
+            this.downPosition = {
+                ...this.downPosition,
+                offsetX: -1,
+                offsetY: mustBePlacedOutsideTheInput ? 10 : -11,
+            };
+        }
```
sergiubologa, line 1299: You can extract this into a private function `_setDropdownOffset`

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
```diff
@@ -1242,6 +1282,40 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
...
+                ...this.upPosition,
+                originX: 'start',
+                originY: 'top',
+            };
+            this.downPosition = {
+                ...this.downPosition,
+                originX: 'start',
+                originY: 'bottom',
+            };
+        }
```
sergiubologa, line 1312: Also, this one can be extracted into a private function `_setDropdownOrigin`

sergiubologa (changes requested):

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
ZoicanDenisUi: Changed it

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
ZoicanDenisUi: Changed it

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
```diff
@@ -1067,6 +1066,8 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
         if (!refocus) {
             this._focusChanged(false);
         }
+
+        if (!this.expandInline) { this._suggestContainerOberserver.unobserve(this.suggestContainer.nativeElement); }
```
sergiubologa, line 1070: I think the check should be like this
```suggestion
        if (!this.suggestContainer) { this._suggestContainerOberserver.unobserve(this.suggestContainer.nativeElement); }
```

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
```diff
@@ -1283,46 +1284,61 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
...
-                originX: 'start',
-                originY: 'bottom',
-            };
+        if (mustBePlacedBesideTheInput) {
+            this._setDropdownOrigin();
         }
 
         this.dropdownPosition = this.direction
             ? [this.direction === 'up' ? this.upPosition : this.downPosition]
             : [this.downPosition];
```
sergiubologa:
```suggestion
        this.dropdownPosition = [this.direction && this.direction === 'up' ? this.upPosition : this.downPosition];
```

projects/angular/components/ui-suggest/src/ui-suggest.component.scss
```diff
@@ -285,58 +120,130 @@ $componentName: "ui-suggest";
...
+            &.is-loading {
+                font-style: italic;
+                opacity: 0.8;
+            }
 
-                &.searchable {
-                    bottom: -17px;
+            &.text-ellipsis:not(.custom-value-template) .mat-list-item-content {
+                display: block;
+            }
```
sergiubologa, line 200: This is causing this alignment issue:
image

sergiubologa: Not sure if this `display: block` is the problem, but there is an issue with the new styles

sergiubologa (changes requested):

projects/angular/components/ui-suggest/src/ui-suggest.component.scss
ZoicanDenisUi: Changed scss

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
ZoicanDenisUi: Changed it

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
ZoicanDenisUi: I removed this condition because `suggestContainer` will never be null because is the root element. I refactored the code and added the codition `!this.isExpandInline` because we need to do this just in this case

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
ZoicanDenisUi: Changed it.

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
ZoicanDenisUi: The `_initDropdownPositions`  will init the positions that will be used in this logic (this.upDirection and this.downPosition). If we call this function again here, the variables will be initialized with the same values. I changed the code and refactored a little for a better understanding.

sergiubologa:
One more issue with the placeholder styling:
image

sergiubologa:
Also, the 2 bottom examples in the playground page don't show the list of elements when first opened:
image

if I close the dropdown and open it up again it loads them.

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
```diff
@@ -868,6 +868,7 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
             .pipe(
                 delay(0),
                 filter(_ => this.isOpen),
+                delay(0),
```
sergiubologa, line 871: Why is this needed? If we move the delay above isn't it enough? What if you use `observeOn(asyncScheduler)`?

sergiubologa: fixed it by using `observeOn(animationFrameScheduler),`

sergiubologa (approved):

sergiubologa (approved):

anbalase (approved):
code looks fine, breaking changes are manageable. pls bump minor and try to state the changes.
cc: @CalinaCristian for apollo overrides

sergiubologa (approved):

PR #316, opened 2023-03-15
feat(grid): add showFirstLastButtons option

dkorobeinikov:


projects/playground/src/app/pages/grid/grid.models.ts
```diff
@@ -29,4 +29,5 @@ export interface IInputs {
     showHeaderRow: boolean;
     customFilter: boolean;
     useCardView: boolean;
+    showFirstLastButtons: boolean;
```
toxik, line 32:
```suggestion
    showFirstLastButtons?: boolean;
```
to preserve backcompat?

adrian-macuc (approved):

PR #320, opened 2023-03-20
Fix(grid): focus active filter item

CatalinB7:


projects/angular/components/ui-grid/src/ui-grid.component.ts
```diff
@@ -1063,6 +1063,11 @@ export class UiGridComponent<T extends IGridDataEntry> extends ResizableGrid<T>
         this.focusedColumnHeader = false;
     }
 
+    focusActiveFilterItem() {
+        const activeItem: HTMLElement | null = document.querySelector('.cdk-overlay-container .active[role="menuitem"]')!;
```
caleniuc, line 1067: If you expect it to be `null` then `!` not needed.

CatalinB7: true, at first i wanted to say it will always find an element having `.cdk-overlay-container .active[role="menuitem"]`, but why risk it

caleniuc (approved):

PR #333, opened 2023-05-09
fix(suggest): condition title display by input

CatalinB7:


projects/angular/components/ui-suggest/src/ui-suggest.mat-form-field.ts
```diff
@@ -103,6 +103,9 @@ export abstract class UiSuggestMatFormFieldDirective implements
         this.stateChanges.next();
     }
 
+    @Input()
```
anbalase, line 106: please add a comment + a test describing what this controls.

anbalase (changes requested):

projects/angular/components/ui-suggest/src/ui-suggest.mat-form-field.ts
anbalase: second guessing this. Maybe we should just expose an aria-label input? wdyt?

PR #339, opened 2023-05-11
feat(suggest): add input for aria-label

CatalinB7:


projects/angular/components/ui-suggest/src/ui-suggest.mat-form-field.ts
```diff
@@ -110,6 +110,13 @@ export abstract class UiSuggestMatFormFieldDirective implements
     @Input()
     shouldHideTitle = false;
 
+    /**
+     * Sets aria-label on input or mat-chip-list element.
+     *
+     */
+    @Input('aria-label')
```
sergiubologa, line 117: I would call it `ariaLabel` to not confuse it with the native attribute

CatalinB7: I was thinking of imitating https://material.angular.io/components/checkbox/api
```@Input('aria-label')
ariaLabel: string```

sergiubologa (approved):

sergiubologa (approved):

PR #345, opened 2023-05-16
feat(a11y): add tabIndex to matIcons with matTooltips

toxik:


projects/angular/a11y/src/ui-auto-accessible-label/ui-auto-accessible-label.directive.ts
```diff
@@ -23,19 +25,32 @@ You can disable this directive using the boolean ${DISABLE_AUTO_ACCESSIBLE_LABEL
...
     matTooltip?: string;
 
     constructor(
-        elementRef: ElementRef,
+        private _elementRef: ElementRef<HTMLUnknownElement>,
         @Optional() tooltip?: MatTooltip,
     ) {
-        if (!tooltip) {
-            console.warn(MAT_TOOLTIP_MISSING_WARNING, elementRef.nativeElement);
+        if (!tooltip && typeof _elementRef.nativeElement.tabIndex !== 'undefined') {
```
petejank, line 39: I think we can drop the `_elementRef.nativeElement.tabIndex !== 'undefined'` check - after all, we still may want to allow the user to provide the `tabindex` manually, since why not.

projects/angular/a11y/src/ui-auto-accessible-label/ui-auto-accessible-label.directive.ts
```diff
@@ -23,19 +25,32 @@ You can disable this directive using the boolean ${DISABLE_AUTO_ACCESSIBLE_LABEL
...
+            console.warn(MAT_TOOLTIP_MISSING_WARNING, _elementRef.nativeElement);
+        }
+    }
+
+    ngOnChanges(changes: SimpleChanges) {
+        if (changes.matTooltip) {
```
petejank, line 45: Was wondering about guard clauses. But Mihai also doesn't have a strong preference for that, so I guess it's optional in our projects. Wdyt?

projects/angular/a11y/src/ui-auto-accessible-label/ui-auto-accessible-label.directive.ts
```diff
@@ -23,19 +25,32 @@ You can disable this directive using the boolean ${DISABLE_AUTO_ACCESSIBLE_LABEL
...
+            const element = this._elementRef.nativeElement;
+            if (
+                element.tagName === 'MAT-ICON'
+                && !!changes.matTooltip.currentValue
+                && typeof element.tabIndex === 'undefined'
```
petejank, line 50: `tabIndex` will always be an integer 🤔 . I think we may want to check here if it's `=== -1`

projects/angular/a11y/src/ui-auto-accessible-label/ui-auto-accessible-label.directive.ts
toxik: that's precisely why the warning would show up, you have a tabbable element and you didn't set a tooltip on it

petejank: So we only need to change the check to `=== -1` then

toxik: but it should also be compared to -1

projects/angular/a11y/src/ui-auto-accessible-label/ui-auto-accessible-label.directive.ts
toxik: don't know if it helps much here

projects/angular/a11y/src/ui-auto-accessible-label/ui-auto-accessible-label.directive.ts
petejank: Khem, my bad, I meant `!== -1` 🤦

projects/angular/a11y/src/ui-auto-accessible-label/ui-auto-accessible-label.directive.ts
petejank: Was more curious what you think about it. Guard clauses look nicer when the linter permits conditional statements as one-liners imho

PR #349, opened 2023-05-17
Fix/accesible remove chip

gheorghitahurmuz:


anbalase (dismissed):
given the current keyboard nav of the mat-chip comp this seems like the best option.

anbalase (approved):

PR #359, opened 2023-05-29
fix(suggest): add item selected output event

ZoicanDenisUi:
**Description**
Added itemSelected output because it is very useful when we use `custom template`. We can use this output to detect when a item is selected and use that event in the logic of the custom template.
I created an example in the `suggest page` where I implemented a suggest component that uses radio-buttons.

image

### DIFF
--- a/projects/angular/components/ui-suggest/src/ui-suggest.component.ts
+++ b/projects/angular/components/ui-suggest/src/ui-suggest.component.ts
@@ -621,6 +621,10 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
     @Output()
     closed = new EventEmitter<void>();
 
+    @Output()
+    // eslint-disable-next-line rxjs/finnish
+    itemSelected = new Subject<ISuggestValue>();
+
     /**
      * Emits when the overlay is displayed (dropdown open).
      *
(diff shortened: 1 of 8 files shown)


projects/angular/components/ui-suggest/src/ui-suggest.component.ts
```diff
@@ -621,6 +621,10 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
     @Output()
     closed = new EventEmitter<void>();
 
+    @Output()
+    // eslint-disable-next-line rxjs/finnish
+    itemSelected = new Subject<ISuggestValue>();
```
sergiubologa, line 626: why not event emitter?
```suggestion
    itemSelected = new EventEmitter<ISuggestValue>();
```

sergiubologa (approved):

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
ZoicanDenisUi: Changed it

sergiubologa (dismissed):

sergiubologa (approved):

anbalase (approved):

PR #373, opened 2023-07-12
fix(grid): a11y on resize column

CalinaCristian:


anbalase:
we should probably handle in apollo theming

PR #374, opened 2023-07-18
fix(grid): use arrow cursor on column header

dan-mirescu:
Use an arrow cursor instead of text select cursor on headers of columns which are not sortable.

https://uipath.atlassian.net/browse/PLT-29992

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# v14.8.7 (2023-07-20)
+* **grid** arrow cursor on grid header text
+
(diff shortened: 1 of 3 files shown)


alexdudescu (approved):
- [ ] Looks like there are conflicts with the changelog meaning someone probably published a new version. You will have to redo the bump commit 
 
- [ ] Since you prepared to also publish a new version don't forget to run `git push --tags`

alexdudescu (approved):

PR #378, opened 2023-07-21
Fix: show border for column resize

dan-mirescu:
Show colored border for column resize instead of a drag icon.

https://uipath.atlassian.net/browse/PLT-29993

### DIFF
--- a/projects/angular/components/ui-grid/src/ui-grid.component.scss
+++ b/projects/angular/components/ui-grid/src/ui-grid.component.scss
@@ -13,13 +13,16 @@ ui-grid {
     $ui-grid-row-horizontal-padding: 24px;
     $ui-feature-cell-width: 50px;
     $ui-row-border-width: 1px;
-    $ui-header-resize-handle-width: 15px;
-    $ui-grid-default-spacing: 5px;
+    $ui-header-resize-border-width: 6px;
+    $ui-header-resize-anchor-width: 24px;
+    $ui-grid-default-spacing: 10px;
     $ui-border-radius: 4px;
     $ui-grid-button-size: 40px;
     $ui-grid-main-font-size: 13.5px;
     $ui-grid-secondary-font-size: 12px;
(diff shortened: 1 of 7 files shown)


gheorghitahurmuz (dismissed):

anbalase (changes requested):
with this change, the new UI communicates the resize anchor a bit too agressively. Pls sync w @alexandramehigh to get an idea if this is acceptable. I'd advise against. If we use this rendering, then the anchor should display/react near column header border and NOT on whole column header.

projects/angular/components/ui-grid/src/ui-grid.component.scss
```diff
@@ -13,7 +13,7 @@ ui-grid {
     $ui-grid-row-horizontal-padding: 24px;
     $ui-feature-cell-width: 50px;
     $ui-row-border-width: 1px;
-    $ui-header-resize-handle-width: 15px;
+    $ui-header-resize-handle-width: 1px;
```
anbalase, line 16: are we making the handle only 1px? seems like a usability concern... pls also check w @alexandramehigh

anbalase: 24px for optimum usability, though we could keep 15 for now if real-estate limitations won't allow it.

dan-mirescu: The 1px was initially added to $ui-grid-default-spacing, resulting in the requested 6 pixels. However, the new changes contain improvements in this area.

anbalase:
no changes pushed

alexdudescu:
image
Having a seamless line from top to bottom when resizing (meaning row bottom borders won't show in between the blue elements) is complicated to pull off from a css standpoint and I think we should defer it. cc: @anbalase Wdyt?

anbalase:
@alexdudescu I don't think we should add these changes to 14. These are just cosmetic changes

PR #391, opened 2023-09-04
fix(suggest): reduce height of ui-suggest when no results found

dan-mirescu:
Reduce the height of ui-suggest when no results are found and customItemSize is used.

https://uipath.atlassian.net/browse/OR-65722


anbalase (approved):
we'll need this for 14 as well, right?

PR #393, opened 2023-09-06
🍒 fix(suggest): reduce height of ui-suggest when no results found

dan-mirescu:
Reduce the height of ui-suggest when no results are found and customItemSize is used.

https://uipath.atlassian.net/browse/OR-65722

### DIFF
--- a/projects/angular/components/ui-suggest/src/ui-suggest.component.html
+++ b/projects/angular/components/ui-suggest/src/ui-suggest.component.html
@@ -304,7 +304,7 @@
             </ng-container>
 
             <mat-list-item *ngIf="loading$ | async"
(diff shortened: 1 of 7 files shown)


projects/angular/components/ui-suggest/src/ui-suggest.component.html
```diff
@@ -304,7 +304,7 @@
             </ng-container>
 
             <mat-list-item *ngIf="loading$ | async"
-                           [style.height.px]="itemSize">
+                           [style.height.px]="baseSize">
```
caleniuc, line 307: If we set the height to `baseSize`, will a custom height set via `customItemSize` modify an item's height?

caleniuc (approved):

PR #394, opened 2023-09-20
🍒 feat(grid): show border for column resize

dan-mirescu:
Show colored border for column resize instead of a drag icon.

https://uipath.atlassian.net/browse/PLT-29993

It is a prerequisite for the horizontally scrollable grid feature.

### DIFF
--- a/projects/angular/components/ui-grid/src/ui-grid.component.scss
+++ b/projects/angular/components/ui-grid/src/ui-grid.component.scss
@@ -13,7 +13,8 @@ ui-grid {
     $ui-grid-row-horizontal-padding: 24px;
     $ui-feature-cell-width: 50px;
     $ui-row-border-width: 1px;
-    $ui-header-resize-handle-width: 15px;
+    $ui-header-resize-handle-width: 6px;
+    $ui-cell-resizing-border-width: 3px;
     $ui-grid-default-spacing: 5px;
     $ui-border-radius: 4px;
     $ui-grid-button-size: 40px;
@@ -134,6 +135,8 @@ ui-grid {
     }
 
     .ui-grid {
+
+
         &-primary,
         &-header-cell {
             font-weight: 500;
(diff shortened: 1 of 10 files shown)


projects/angular/components/ui-grid/src/ui-grid.component.scss
```diff
@@ -365,10 +368,11 @@ ui-grid {
             display: block;
 
             .ui-grid-header-title {
-                width: calc(100% - #{$ui-header-resize-handle-width + $ui-grid-default-spacing});
+                width: 100%;
                 height: 100%;
                 display: inline-flex;
                 align-items: center;
+                position: relative;
```
gheorghitahurmuz, line 375: why do we need position relative?

projects/angular/components/ui-grid/src/ui-grid.component.scss
```diff
@@ -438,6 +442,23 @@ ui-grid {
                 }
             }
 
+            .ui-grid-cell-resizing-border {
+                flex: 0 0 $ui-cell-resizing-border-width;
+                height: $ui-grid-row-height + $ui-row-border-width;
+                margin-top: -$ui-row-border-width;
```
gheorghitahurmuz, line 448: nit: negative values like `-1 * $ui-row-border-width` may be more readable

gheorghitahurmuz (approved):

projects/angular/components/ui-grid/src/ui-grid.component.scss
```diff
@@ -438,6 +442,23 @@ ui-grid {
...
+            }
+
+            .ui-grid-header-cell {
+                &.ui-grid-spacer-cell {
+                    min-width: $ui-feature-cell-width;
+                }
+            }
+
+            .ui-grid-header-cell {
+                justify-content: space-between;
```
caleniuc, line 459: In what case do we need this?

nit: can be placed inside the above rule, just for `.ui-grid-header-cell`

caleniuc (approved):

projects/angular/components/ui-grid/src/ui-grid.component.scss
dan-mirescu: It seems it's not needed. I will remove it.

PR #411, opened 2023-10-23
Fix(grid): limit sticky columns max width  

CatalinB7:
Todo:
- [x] fix / add tests

### DIFF
--- a/projects/angular/components/ui-grid/src/managers/resize/resize-manager.ts
+++ b/projects/angular/components/ui-grid/src/managers/resize/resize-manager.ts
@@ -330,8 +330,8 @@ export abstract class ResizeManager<T extends IGridDataEntry> {
         if (!this._definitions?.length) {
             return 0;
         }
-        const totalColumnWidths = this._definitions.reduce((acc, curr) => acc + +curr.width, 0);
-        const totalHeaderWidths = this._headers!.reduce((acc, curr) => acc + curr.getBoundingClientRect().width, 0);
+        const totalColumnWidths = this._definitions.reduce((acc, curr) => acc + Number(curr.width), 0);
+        const totalHeaderWidths = this._headers!.reduce((acc, curr) => acc + curr.getBoundingClientRect().width, 0) || 1;
 
         return totalColumnWidths / totalHeaderWidths;
     }
--- a/projects/angular/components/ui-grid/src/managers/resize/resize-manager.ts
+++ b/projects/angular/components/ui-grid/src/managers/resize/resize-manager.ts
@@ -89,13 +89,14 @@ export abstract class ResizeManager<T extends IGridDataEntry> {
     protected _headers?: HTMLDivElement[];
     protected _table?: HTMLTableElement | null;
     protected _definitions?: UiGridColumnDirective<T>[];
+    protected _gridElement: HTMLDivElement;
+    protected _gridResizeObserver?: ResizeObserver;
+    protected _widthMap = new Map<string, number>();
+
     private _previous?: IResizeState<T> = {} as IResizeState<T>;
     private _direction?: ResizeDirection;
     private _resize$ = new Subject<IResizeEvent<T>>();
     private _stopped$ = new Subject<void>();
-    private _widthMap = new Map<string, number>();
-    private _gridElement: HTMLDivElement;
-    private _gridResizeObserver?: ResizeObserver;
     private _widthChange$: Subject<number> = new Subject();
 
     // eslint-disable-next-line @typescript-eslint/member-ordering
@@ -104,7 +105,7 @@ export abstract class ResizeManager<T extends IGridDataEntry> {
     );
 
     constructor(
-        private _grid: ResizableGrid<T>,
+        protected _grid: ResizableGrid<T>,
     ) {
         // eslint-disable-next-line no-underscore-dangle
         this._gridElement = (_grid as any)._ref.nativeElement;
@@ -325,6 +326,16 @@ export abstract class ResizeManager<T extends IGridDataEntry> {
         }
     }
 
+    protected _computePixelsToPercentRatio() {
+        if (!this._definitions?.length) {
+            return 0;
+        }
+        const totalColumnWidths = this._definitions.reduce((acc, curr) => acc + +curr.width, 0);
+        const totalHeaderWidths = this._headers!.reduce((acc, curr) => acc + curr.getBoundingClientRect().width, 0);
+
+        return totalColumnWidths / totalHeaderWidths;
+    }
+
     private _getCellsFor = (property: string) => toArray<HTMLDivElement>(
         this._gridElement
             .querySelectorAll(cellSelector(property)),
@@ -352,15 +363,5 @@ export abstract class ResizeManager<T extends IGridDataEntry> {
             }
         });
         this._previous = {} as IResizeState<T>;
-    }
-
-    private _computePixelsToPercentRatio() {
-        if (!this._definitions?.length) {
-            return 0;
-        }
-        const totalColumnWidths = this._definitions.reduce((acc, curr) => acc + +curr.width, 0);
-        const totalHeaderWidths = this._headers!.reduce((acc, curr) => acc + curr.getBoundingClientRect().width, 0);
-
-        return totalColumnWidths / totalHeaderWidths;
     }
 }
--- a/projects/angular/components/ui-grid/src/managers/resize/strategies/scrollable-grid-resizer.ts
+++ b/projects/angular/components/ui-grid/src/managers/resize/strategies/scrollable-grid-resizer.ts
@@ -1,7 +1,28 @@
 import { IGridDataEntry } from '../../../models';
 import { ImmediateNeighbourHaltResizer } from './immediate-neighbour-halt-resizer';
 
+export const MAXIMUM_STICKY_COVERAGE = 0.7;
 export class ScrollableGridResizer<T extends IGridDataEntry> extends ImmediateNeighbourHaltResizer<T> {
-    // TODO: check if sticky columns exceeded width limit
+    limitStickyWidthCoverage(tableContainerWidth: number) {
+        const stickyContainerWidth = this._gridElement!.querySelector('.sticky-columns-header-container')!.getBoundingClientRect().width;
+        const exceedingStickyCoverageLimit = stickyContainerWidth / tableContainerWidth > MAXIMUM_STICKY_COVERAGE;
+
+        if (exceedingStickyCoverageLimit) {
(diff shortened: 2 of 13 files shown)


anbalase:
don't forget to cherry once done

projects/angular/components/ui-grid/src/managers/resize/resize-manager.ts
```diff
@@ -325,6 +326,16 @@ export abstract class ResizeManager<T extends IGridDataEntry> {
...
     }
 
+    protected _computePixelsToPercentRatio() {
+        if (!this._definitions?.length) {
+            return 0;
+        }
+        const totalColumnWidths = this._definitions.reduce((acc, curr) => acc + +curr.width, 0);
+        const totalHeaderWidths = this._headers!.reduce((acc, curr) => acc + curr.getBoundingClientRect().width, 0);
+
+        return totalColumnWidths / totalHeaderWidths;
```
gheorghitahurmuz, line 336: could `totalHeaderWidths` be 0?

projects/angular/components/ui-grid/src/managers/resize/strategies/scrollable-grid-resizer.ts
```diff
@@ -1,7 +1,117 @@
...
+    isSticky,
+} from '../resize-manager.constants';
 import { ImmediateNeighbourHaltResizer } from './immediate-neighbour-halt-resizer';
 
+export const MAXIMUM_STICKY_COVERAGE = 0.7;
 export class ScrollableGridResizer<T extends IGridDataEntry> extends ImmediateNeighbourHaltResizer<T> {
-    // TODO: check if sticky columns exceeded width limit
+    limitStickyWidthCoverage(tableContainerWidth: number) {
+        const stickyContainerWidth = this._gridElement!.querySelector('.sticky-columns-header-container')!.getBoundingClientRect().width;
+        const exceedingStickyCoverageLimit = stickyContainerWidth / tableContainerWidth > MAXIMUM_STICKY_COVERAGE;
```
gheorghitahurmuz, line 15: could `tableContainerWidth` be 0?

projects/angular/components/ui-grid/src/managers/resize/strategies/scrollable-grid-resizer.ts
```diff
@@ -1,7 +1,117 @@
...
+    };
+
+    private _isLastStickyColumn(state: IResizeEvent<T>) {
+        return state.current.resized.element.classList.contains('ui-grid-sticky-element') &&
+            !state.current.neighbour?.element.classList.contains('ui-grid-sticky-element');
+    }
+
+    private _limitOffsetForStickyContainer(offset: number) {
+        const ratio = this._computePixelsToPercentRatio();
+        const stickyColumnsPercentageSum = this._grid.columns.filter(c => c.isSticky).map(c => +c.width).reduce((a, b) => a + b, 0);
```
gheorghitahurmuz, line 89: c -> column, a -> sum, b -> columnWidth

projects/angular/components/ui-grid/src/ui-grid.component.html
```diff
@@ -204,18 +204,16 @@
...
                                             </div>
                                         </div>
-                                    </div>
-
-                                    <div *ngIf="!isProjected && column.resizeable && !useCardView && (!last || columns.freeColumns.length > 0)"
-                                        [class.last-sticky-handle-container]="last && columns.freeColumns.length > 0"
-                                         class="ui-grid-resize-handle-container ui-grid-sticky-element">
 
-                                        <div [class.ui-grid-state-resizing]="column === resizeManager.current?.column"
+                                        <div *ngIf="!isProjected && column.resizeable && !useCardView && (!last || columns.freeColumns.length > 0)"
```
gheorghitahurmuz, line 208: I think there's too much logic in this template

projects/angular/components/ui-grid/src/ui-grid.component.html
```diff
@@ -411,7 +408,7 @@
          *ngIf="!isRowExpanded(row?.id) || expandMode === 'preserve'"
          [class.ui-grid-row-state-expanded]="isRowExpanded(row?.id)"
          [class.ui-grid-border-none]="!footer && last"
-         [class.highlighted-row]="highlightedEntityId === row?.id"
+         [class.highlighted-row]="highlightedEntityId != null && highlightedEntityId.toString() === row.id.toString()"
```
gheorghitahurmuz, line 411: why `toString()`?

projects/angular/components/ui-grid/src/ui-grid.component.scss
```diff
@@ -468,6 +469,10 @@ ui-grid {
                 }
 
                 &.highlighted-row {
+                    .ui-grid-cell, .ui-grid-cell-resizing-border-container {
+                        background-color:$ui-grid-row-hover-color;
```
gheorghitahurmuz, line 473:
```suggestion
                        background-color: $ui-grid-row-hover-color;
```

projects/angular/components/ui-grid/src/ui-grid.component.scss
```diff
@@ -555,6 +538,44 @@ ui-grid {
...
+
+                .ui-grid-header-cell:not(.ui-grid-checkbox-cell) {
+                    position: relative !important;
+                }
+
+                .ui-grid-resize-handle-container {
+                    position: absolute;
+                    right: 0;
+                    top: 0;
+                    width: 18px;
```
gheorghitahurmuz, line 553: extract these value to named variables

projects/angular/components/ui-grid/src/ui-grid.component.scss
```diff
@@ -555,6 +538,44 @@ ui-grid {
...
+                    padding-left: 12px;
+                    box-sizing: border-box;
+                    left: unset !important;
+                    background-color: transparent;
+
+                    cursor: col-resize;
+
+                    .ui-grid-resize-handle {
+                        width: 6px;
```
gheorghitahurmuz, line 562: also here

projects/angular/components/ui-grid/src/ui-grid.component.ts
```diff
@@ -817,6 +818,7 @@ export class UiGridComponent<T extends IGridDataEntry>
     private _isShiftPressed = false;
     private _lastCheckboxIdx = 0;
     private _resizeSubscription$: null | Subscription = null;
+    private _containerWidthChangeSubscription$: null | Subscription = null;
```
gheorghitahurmuz, line 821:
```suggestion
    private _containerWidthChange$: null | Subscription = null;
```

projects/angular/components/ui-grid/src/ui-grid.component.ts
```diff
@@ -1232,6 +1235,16 @@ export class UiGridComponent<T extends IGridDataEntry>
 
             this.resizeEnd.emit();
         });
+
+        if (this.isScrollable) {
+            this._containerWidthChangeSubscription$ = this.resizeManager.widthChange$.pipe(
+                distinctUntilChanged(),
+                tap(width => {
+                    (this.resizeManager as ScrollableGridResizer<T>).limitStickyWidthCoverage(width);
```
gheorghitahurmuz, line 1243: could we avoid type casting?

projects/angular/components/ui-grid/src/ui-grid.component.html
alexdudescu: That's because we can have either ids (numbers) or guids(strings) so in order to accommodate for both we cast to string.

projects/angular/components/ui-grid/src/ui-grid.component.html
alexdudescu: I agree on this topic but in order to be in time with the delivery I suggest we address this issue as a general angular components cleanup tasks. The suggest suffers of the same single file functionality cramming

projects/angular/components/ui-grid/src/ui-grid.component.html
CatalinB7: the consumer (OR) sets the highlightedEntityId as string instead of number because the value is retrieved from [ParamMap](https://angular.io/api/router/ParamMap). It seems simpler to just convert the values to strings when comparing

projects/angular/components/ui-grid/src/ui-grid.component.ts
CatalinB7: i don't think so. resizeManager is of type `ResizeManager` and `ScrollableGridResizer` extends this class and adds the method `limitStickyWidthCoverage`. One way to solve this would be to add `limitStickyWidthCoverage` on  the base class but i don't think that would be correct to do since the method is relevant only in scrollable grid context.

projects/angular/components/ui-grid/src/ui-grid.component.ts
CatalinB7: i would stick to the pattern on the line above (`private _resizeSubscription$: null | Subscription = null;`)

projects/angular/components/ui-grid/src/managers/resize/strategies/scrollable-grid-resizer.ts
CatalinB7: added a fallback to 1

projects/angular/components/ui-grid/src/managers/resize/strategies/scrollable-grid-resizer.ts
```diff
@@ -1,7 +1,117 @@
...
+        if ((stickyContainerWidth + offsetPx) / tableWidth > MAXIMUM_STICKY_COVERAGE) {
+            offset = Math.max((tableWidth * MAXIMUM_STICKY_COVERAGE - stickyContainerWidth) * ratio, 0);
+        }
+        return offset;
+    }
+
+    private _isExceedingStickyCoverageLimit() {
+        const stickyContainer = this._table!.querySelector('.sticky-columns-header-container');
+        const stickyContainerWidth = stickyContainer!.getBoundingClientRect().width;
+        const tableContainerWidth = this._table!.getBoundingClientRect().width;
```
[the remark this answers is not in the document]
CatalinB7: created `elementWidth` instead to be a bit more generic

projects/angular/components/ui-grid/src/managers/resize/strategies/scrollable-grid-resizer.ts
CatalinB7: applied suggestion. but i think one letter names were good enough

["LOCAL variable names should be short, and to the point. If you have some random integer loop counter, it should probably be called i. Calling it loop_counter is non-productive, if there is no chance of it being mis-understood. Similarly, tmp can be just about any type of variable that is used to hold a temporary value."](https://www.kernel.org/doc/html/v4.10/process/coding-style.html#naming)

PR #421, opened 2023-11-03
Fix/grid a11y

CatalinB7:
OR-68919


alexdudescu:
There's one failing test: 
```
Component: UiGrid Scenario: simple grid Configuration: grid with columns, rows, refresh and checkboxes State: populated Feature: radio button selection should display Select / Deselect according to button state
```

alexdudescu (approved):

PR #428, opened 2023-11-14
fix: lazy suggest buffer size

Gabriela-David:
# Jira

[SOL-1407](https://uipath.atlassian.net/browse/SOL-1407)

[SOL-1407]: https://uipath.atlassian.net/browse/SOL-1407?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/projects/angular/components/ui-suggest/src/ui-suggest.component.ts
+++ b/projects/angular/components/ui-suggest/src/ui-suggest.component.ts
@@ -474,6 +474,10 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
         return itemsLength + (this._hasCustomValue$.value && !this.isDown ? 0 : -1);
     }
 
+    private get _fetchCount() {
+        return this._isLazyMode ? this.displayCount * 2 : this.displayCount;
+    }
+
     /**
      * A search stream factory, generally used to retrieve data from the server when a user searches.
      * By `default`, a search factory is generated that does an `in-memory` lookup if `searchable` is set to `true`.
@@ -719,13 +723,6 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
                 takeUntil(this._destroyed$),
             )
             .subscribe(start => {
-                if (this._isLazyMode
-                    && this.isDown
-                    && this._items[start + this.displayCount]?.loading === VirtualScrollItemStatus.initial) {
-
-                    this._loadMore();
-
-                }
                 this._visibleRange = {
                     start,
                     end: start + this.displayCount,
@@ -1275,8 +1272,7 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
         this.loading$.next(true);
 
         if (this._searchSub) { this._searchSub.unsubscribe(); }
-
-        this._searchSub = this.searchSourceFactory(searchValue, this.displayCount)
+        this._searchSub = this.searchSourceFactory(searchValue, this._fetchCount)
             .pipe(
                 tap(this._setInitialItems),
                 map(this._findItemIndex(searchValue)),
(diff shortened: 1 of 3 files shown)


anbalase:
since this is an improvement, I recommend we only deliver to master

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
```diff
@@ -719,13 +723,6 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
                 takeUntil(this._destroyed$),
             )
             .subscribe(start => {
-                if (this._isLazyMode
```
anbalase, line 722: why was this removed?

anbalase:
@gheorghitahurmuz can you also pls review? you have good knowledge of this strategy.

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
Gabriela-David: `_loadMore` calls `rangeLoad` and with the way lazy suggest works now, `rangeLoad` would be called twice when it is only needed once

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
```diff
@@ -1752,16 +1746,16 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
     }
 
     private _shouldAddLoadingElementInLazyMode(currentResponse: ISuggestValue[]) {
-        return this._isLazyMode && currentResponse.length >= this.displayCount;
+                return this._isLazyMode && currentResponse.length >= this.displayCount;
     }
 
     private _addLoadingElementInLazyMode() {
-        const loadingElement = generateLoadingInitialCollection(this.intl.loadingLabel, 1)[0];
+        const loadingElement = generateLoadingInitialCollection(this.intl.loadingLabel, this._fetchCount);
```
gheorghitahurmuz, line 1753: I believe there are several unused loading elements. For instance, we generate 20, but receive only 0 or 1 element from the backend. As a result, the remaining elements (e.g., 19 out of 20) are removed, which will affect the scrollbar.

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
```diff
@@ -1752,16 +1746,16 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
...
 
         if (this.isDown) {
-            this._items.push(loadingElement);
+            this._items = [...this._items, ...loadingElement];
```
gheorghitahurmuz, line 1756: why not `this._items.push(...loadingElement)`?

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
```diff
@@ -1752,16 +1746,16 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
...
         } else {
-            this._items.unshift(loadingElement);
+            this._items = [ ...loadingElement, ...this._items];
```
gheorghitahurmuz, line 1758: same ^

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
```diff
@@ -1655,7 +1649,7 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
                     ),
                 ),
                 finalize(
-                    () => resetUnloadedState(this._items, mappedStart, mappedEnd),
+                    () => resetUnloadedState(this._items, mappedStart, mappedEnd, this._isLazyMode),
```
gheorghitahurmuz, line 1652: I think we can implement the condition here and remove the flag related to lazy loading

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
```diff
@@ -1629,14 +1625,12 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
 
         this._lazyLoadLastArgument = newArguments;
 
-        const mappedStart = this.isDown ?
-            this._isLazyMode ? this._items.length - 1 : start :
+        const mappedStart = this.isDown ? start :
             this._isLazyMode ? -1 : this._items.length - end - 1;
```
gheorghitahurmuz, line 1629: If you have the condition `isLazy ? 0 : mappedStart, isLazy ? items.length : mappedEnd` implemented below, then you should modify it to `const mappedStart = this.isDown ? start : this._items.length - end - 1` here

PR #438, opened 2023-12-08
feat(file-picker): file dropzone directive

ioanajoj:
Split file-drop-zone.component into 2
  - directive that handles file drag and drop on any kind of component
  - component that extends the directive and adds an input to also allow opening the browser file picker


sergiubologa (dismissed):

projects/angular/components/ui-file-picker/src/shared/file-drop-zone/file-drop-zone.directive.ts
```diff
@@ -0,0 +1,131 @@
...
+import {
+ Subject, map, takeUntil,
+} from 'rxjs';
+import { FileReaderService } from '../file-reader.service';
+import { sortAndFilter } from '../file-drop-zone.utils';
+
+// The highlight class gets added to the host HTML element,
+// so to display properly the component containing it
+// should have that class defined in its styles.
+const DROP_ZONE_HIGHLIGHT_CLASS = 'file-drop-zone-highlight';
```
sergiubologa, line 14: I would rename it like this
```suggestion
const DROP_ZONE_HIGHLIGHT_CLASS = 'ui-file-drop-zone-highlight';
```

projects/angular/components/ui-file-picker/src/shared/file-drop-zone/file-drop-zone.directive.ts
```diff
@@ -0,0 +1,131 @@
...
+        this._fileReaderService.error$
+            .pipe(
+                takeUntil(this._destroyed$),
+            )
+            .subscribe(err => this.fileError.emit(err));
+    }
+
+    ngAfterViewInit() {
+        if (!this.dropZone) {
+            this.dropZone = this._element.nativeElement;
```
sergiubologa, line 108: Do you need this property? Also, why is it public?

sergiubologa (approved):

PR #444, opened 2023-12-20
fix(grid-visibility-manager): consider only checked columns when computing `isDirty$`

CatalinB7:


AlexAndriesUiPath (approved):

projects/angular/components/ui-grid/src/managers/visibility-manager.ts
```diff
@@ -37,7 +37,7 @@ export class VisibilityManger<T extends IGridDataEntry> {
     isDirty$ = this.options$.pipe(
         filter(() => !!this._initial),
         map(o => ([o.map(this._mapToVisibleDiff), this._initial])),
-        map(([current, initial]) => !isEqual(current, initial)),
+        map(([current, initial]) => !isEqual(current?.filter(v => v.checked), initial?.filter(v => v.checked))),
```
anbalase, line 40: if I start with a,b, (c off) and check c it should mark as dirty and not ignore it, right?

anbalase: nevermind

anbalase (dismissed):

AlexAndriesUiPath (approved):

PR #446, opened 2023-12-28
fix: snackbar apply icon color

ivnbogdan:
Consume icon-color coming from theme, and apply it based on snackbar level

Before:
image

After:
error
image

info
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# v15.1.1 (2023-12-28)
+* **snackbar** apply icon-color
+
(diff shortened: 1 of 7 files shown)


toxik (approved):
very nice

ivnbogdan:
This is already handled in apollo here: https://github.com/UiPath/apollo-design-system/blob/master/packages/apollo-angular-mdc-theme/src/custom-overrides/components/_ui-snackbar-apollo.theme.scss

Consuming code should be loading the `angular-components-overrides` mixin

PR #449, opened 2024-01-22
feat(ng-let): support typings

robertandreimoldoveanu:
before:
image

after:
image


anbalase (approved):
do the bump

robertandreimoldoveanu:
closing for now as it creates a breaking change. will either add it once we migrate or as gradual rollout (one directive with typings and the `OldNgLetDirective` without

PR #453, opened 2024-01-31
fix(file-picker): dropzone should only highlight when dragging files

ioanajoj:
## Description

- do not highlight drag area when dragging event does not contain Files

Before
file-picker-disable-drag-anchors-before

After
file-picker-disable-drag-anchors-after


sergiubologa (dismissed):
can this scenario be tested with a unit test?

sergiubologa (dismissed):

sergiubologa (approved):

PR #454, opened 2024-02-05
Add support for High Density mode in ui-grid

eminecula07:
https://uipath.atlassian.net/wiki/spaces/OR/pages/87319554094/High+Density+Layout

[OR-75788](https://uipath.atlassian.net/browse/OR-75788)

[OR-75788]: https://uipath.atlassian.net/browse/OR-75788?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


projects/angular/components/ui-grid/src/ui-grid.component.ts
```diff
@@ -345,7 +354,12 @@ export class UiGridComponent<T extends IGridDataEntry>
      *
      */
     @Input()
-    rowSize: number;
+    get rowSize() {
+        return this._rowSize;
+    }
+    set rowSize(value: number) {
+        this._rowSize = value;
+    }
```
CatalinB7, line 362: is this setter/getter actually needed?

eminecula07: rowSize it's an input and we want to change it in ngOnChange; so I think we need this

AlexAndriesUiPath: unless you do something else when `rowSize` input is reassigned, you don't need a setter/getter

`@Input() rowSize!: number;` is the same thing as what you have

projects/angular/components/ui-grid/src/ui-grid.component.ts
```diff
@@ -1065,6 +1082,13 @@ export class UiGridComponent<T extends IGridDataEntry>
             this._performanceMonitor.reset();
             this.selectionManager.clear();
         }
+
+        const hasHighDensityChange = changes.hasHighDensity;
+        if (hasHighDensityChange?.currentValue) {
```
AlexAndriesUiPath, line 1087: what happen if the `currentValue` is `false`?

eminecula07: Oh, you're right! I'll update! Thanks!

projects/angular/components/ui-grid/src/ui-grid.component.ts
CatalinB7: the code in `ngOnChange` could change `rowSize` directly, i don't see the need for `_rowSize`

AlexAndriesUiPath: the property is changed also by a flag `hasHighDensity`
however, the setter should also check if the flag is enabled to enforce the right `rowSize`

projects/angular/components/ui-grid/src/ui-grid.component.ts
```diff
@@ -1065,6 +1082,13 @@ export class UiGridComponent<T extends IGridDataEntry>
             this._performanceMonitor.reset();
             this.selectionManager.clear();
         }
+
+        const hasHighDensityChange = changes.hasHighDensity;
```
AlexAndriesUiPath, line 1086: I don't think you need the variable here, doesn't add that much readability

eminecula07: Ok. I just followed the pattern above ...

projects/angular/components/ui-grid/src/ui-grid.component.ts
eminecula07: I made a little app to test if Angular throws any warnings/errors in this kind of situations and it seems it doesn't. I guess I can get rid of _rowSize and keep only rowSize.

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
```diff
@@ -184,6 +184,14 @@ export class UiSuggestComponent extends UiSuggestMatFormFieldDirective
         this._cd.detectChanges();
     }
 
+    /**
+     * Set the grid in high density state.
```
anbalase, line 188: changing the rowHeight may have impact on virtualization of items in range loading scenarios, pls test

eminecula07: It does. That's why I made some changes for rowSize.

projects/angular/components/ui-grid/src/ui-grid.component.ts
eminecula07: Done.

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
anbalase: how does itemSize input impact your solution?
```
/**
     * Set a custom size for the list items.
     *
     */
    @Input()
    customItemSize?: number;

    /**
     * Computes the component item height depending on the current render mode.
     *
     */
    get itemSize() {
        if (this.customItemSize) { return this.customItemSize; }

        return this.baseSize;
    }

    get baseSize() {
        return this.isFormControl ? 32 : 40;
    }

```

anbalase: also the comment is regarding the grid, pls update.

projects/angular/components/ui-grid/src/ui-grid.component.ts
```diff
@@ -899,7 +909,8 @@ export class UiGridComponent<T extends IGridDataEntry>
 
         this.disableSelectionByEntry = () => null;
         this._fetchStrategy = _gridOptions?.fetchStrategy ?? 'onOpen';
-        this.rowSize = _gridOptions?.rowSize ?? DEFAULT_VIRTUAL_SCROLL_ITEM_SIZE;
+        this.hasHighDensity = Boolean(this._gridOptions?.hasHighDensity);
```
anbalase, line 912: nit: try to be consistent when a pattern is available already.
```suggestion
        this.hasHighDensity = this._gridOptions?.hasHighDensity ?? false;
```

projects/angular/components/ui-grid/src/ui-grid.component.ts
```diff
@@ -190,6 +192,14 @@ export class UiGridComponent<T extends IGridDataEntry>
     @Input()
     isProjected: boolean;
 
+    /**
+     * Set the grid in high density state.
+     *
+     */
+    @HostBinding('class.ui-grid-state-high-density')
+    @Input()
+    hasHighDensity = false;
```
anbalase, line 201: pls add tests for the following scenarios:
- default - should NOT be in highDensity mode
- set by token - should be able to set highDensity mode via token
- set by input over token - should be able to override highDensity mode via input if token is added.

checks should look at rendered class and rowSize

projects/angular/components/ui-suggest/src/ui-suggest.component.scss
```diff
@@ -13,6 +12,14 @@ $ui-suggest-focus-transition: background-color 200ms cubic-bezier(0.35, 0, 0.25,
 
 $componentName: "ui-suggest";
 
+ui-suggest {
```
anbalase, line 15: pls reuse existing .scss structure. You already have `#{$componentName} {`

projects/angular/components/ui-grid/src/models/dataModel.ts
```diff
@@ -30,4 +30,5 @@ export interface GridOptions<T> {
     rowSize?: number;
     resizeStrategy?: ResizeStrategy;
     selectablePageIndex?: boolean;
+    hasHighDensity?: boolean;
```
anbalase, line 33: how do rowSize & hasHighDensity inputs make sense together?
if we are confused, how would we explain to  consumers know how to configure this?

projects/angular/components/ui-grid/src/ui-grid.component.ts
```diff
@@ -1,3 +1,4 @@
+import { isArray } from 'lodash-es';
```
anbalase, line 1: nit:
```suggestion
import  isArray from 'lodash-es/isArray';
```

projects/angular/components/ui-grid/src/ui-grid.component.scss
```diff
@@ -25,7 +25,7 @@ ui-grid {
     --ui-grid-secondary-font-size: 12px;
     --ui-grid-action-buttons-spacing: 16px;
 
-    &.ui-grid-state-high-density.ui-grid-state-high-density.ui-grid-state-high-density {
+    &.ui-grid-state-high-density {
```
anbalase, line 28: this commit should be a fixup. since changes have not yet reached master I suggest you compact your commits

projects/angular/components/ui-grid/src/ui-grid.component.scss
```diff
@@ -275,7 +275,7 @@ ui-grid {
 
             .mat-mdc-paginator {
                 &-container {
-                    min-height: calc(var(--ui-grid-header-row-height) - 1);
+                    min-height: calc(var(--ui-grid-header-row-height) - 1px);
```
anbalase, line 278: again, fixup

projects/angular/components/ui-grid/src/ui-grid.component.ts
```diff
@@ -1094,9 +1094,8 @@ export class UiGridComponent<T extends IGridDataEntry>
             this.selectionManager.clear();
         }
 
-        const hasHighDensityChange = changes.hasHighDensity;
-        if (hasHighDensityChange?.currentValue) {
-            this._rowSize = hasHighDensityChange.currentValue ?
+        if (changes.hasHighDensity?.currentValue !== undefined) {
+            this._rowSize = changes.hasHighDensity.currentValue ?
```
anbalase: fixup

projects/angular/components/ui-grid/src/ui-grid.component.scss
```diff
@@ -240,6 +240,25 @@ ui-grid {
                 display: flex;
                 margin-left: auto;
                 align-items: center;
+                overflow: hidden;
+
+                .ui-grid-action-buttons {
+                    overflow: hidden;
+
+                    button {
+                        min-width: 88px;
```
anbalase: larger, generic topic here: should these changes not reside here? maybe it should be part of main theme overrides / each consumer. 🤔

anbalase: cc: @alexdudescu @caleniuc wdyt?

alexdudescu: Maybe the answer is somewhere in the middle? (eg. expose some token users cand override that defaults to sth generic.)

projects/angular/components/ui-grid/src/ui-grid.component.scss
```diff
@@ -240,6 +240,25 @@ ui-grid {
                 display: flex;
                 margin-left: auto;
                 align-items: center;
+                overflow: hidden;
+
+                .ui-grid-action-buttons {
+                    overflow: hidden;
+
+                    button {
```
caleniuc: If there are any icon buttons in the rhs side, this would apply to them too. Also, anchor buttons might be used (we have a few cases), where the element's tag is `a`, it wouldn't apply to them. 

```suggestion
                    .mdc-button {
```

projects/angular/components/ui-grid/src/ui-grid.component.scss
anbalase: Ok, regardless of the override solution. Should this be only here?

caleniuc: Not sure if we are talking just in the context of grid action buttons or all the buttons in the app. 

If we want to change the current min width of all buttons, then the overrides are better to be done in the main theme overrides. There is currently a min width set from material of `64px`.

If we want to apply it in the context of the grid, then I think here are ok. When the project will be part of the main theming repo, I don't see a reason to have overrides for these components, instead of updating their default styles to match our theme.

projects/angular/components/ui-grid/src/ui-grid.component.scss
eminecula07: First, I reverted this commit, so it doesn't matter for now.
Second, regarding the anchor tag, I agree; I noticed that while testing and I updated the code here but because I rolled back these changes they it didn't reach git remote. Also, not sure I get what you mean with the button part. Does an icon button not contain a button element?

projects/angular/components/ui-grid/src/ui-grid.component.scss
eminecula07: First, these lines of code were removed in a later commit so this thread should not impact the current PR. However, we have another task for this, so I see the discussion useful,
1. I have a feeling that we should have the same min-width for all the buttons in the app; not sure about this but probably for consistency this is the best option.
2. @caleniuc Regardint the min-width for icon-only buttons, one one hand I agree with you (I think they look better when they're smaller); but yesterday I read an article from @gheorghitahurmuz (https://defensivecss.dev/tip/button-min-width/) that seems to encourage wider buttons (the context is a little different, but I think the same applies in our case: "From a UX perspective, this isn't good as a call to action button must be large enough, specially for touch."). It's just a thing we should think about; we don't focus that much on accessibility and we might be ok with smaller buttons, but we should check with the UX at least.
3. @caleniuc Is 120px a min-width or a general used width? I guess they don't consider all the cases we do (like small window).
4. I don't know why Material Design elements are not very customizable, but I guess we can have a CSS variable one can set to override the min-width and just use the default value to 88px or whatever the best value is (I got it from Dana Moldovan, but I will start a thread with the UX team to figure out the best solution for our issue).

projects/angular/components/ui-grid/src/models/dataModel.ts
eminecula07: hasHighDensity only impacts the default value of rowSize; if one provides a custom value for this input then the default will not be used anyway so it's irrelevant whether we're in high density mode or not.

projects/angular/components/ui-suggest/src/ui-suggest.component.ts
eminecula07: Just for the record, I will update the code so that the list isn't impacted; it's a focused task and we should not apply density for it (as indicated here: https://m2.material.io/design/layout/applying-density.html#usage). Thank you!

projects/angular/components/ui-grid/src/models/dataModel.ts
eminecula07: Update: I checked the code and it seems I forgot we update rowSize when hasHighDensity input is changed. I don't think we need this; at least for our use case (when we set the density mode in a certain page or we toggle a FF) we don't need this. I added this functionality to make the component more independent on its usage, but I think we can remove it.
An alternative solution is to set a private variable like _hasCustomRowSize that we set to true if the input value or the value of rowSize on gridOptions is defined.
I still think it's unnecessary, but maybe you have another opinion!?

projects/angular/components/ui-grid/src/ui-grid.component.spec.ts
```diff
@@ -1059,6 +1060,104 @@ describe('Component: UiGrid', () => {
...
+            });
+
+            fixture = TestBed.createComponent(TestFixtureGridWithHighDensityModeComponent);
+            fixture.detectChanges();
+
+            const grid = fixture.componentInstance.grid;
+            const gridElement = fixture.debugElement.queryAll(By.css('ui-grid'))[0];
+
+            expect(Object.keys(gridElement.classes).includes('ui-grid-state-high-density')).toBeTruthy();
+            expect(grid.rowSize).toBe(32);
```
CatalinB7, line 1134: nit: maybe check the actual height of the div element that represents the row

projects/angular/components/ui-grid/src/ui-grid.component.ts
```diff
@@ -1380,4 +1393,10 @@ export class UiGridComponent<T extends IGridDataEntry>
             }
         }
     }
+
+    private _setInitialRowSize() {
+        this.rowSize = this._gridOptions?.rowSize ?? (this.hasHighDensity ?
+            DEFAULT_VIRTUAL_SCROLL_HIGH_DENSITY_ITEM_SIZE :
+            DEFAULT_VIRTUAL_SCROLL_ITEM_SIZE);
+    }
```
CatalinB7, line 1401: if we want to react on changes to highDensity input then `_setInitialRowSize`'s code should happen in the input setter. That way `ngOnInit` would not be needed.

projects/angular/components/ui-grid/src/ui-grid.component.spec.ts
eminecula07: Done.

projects/angular/components/ui-grid/src/ui-grid.component.ts
eminecula07: What is your suggestion for this?

CatalinB7: `this.rowSize = this._gridOptions?.rowSize ?? DEFAULT_VIRTUAL_SCROLL_ITEM_SIZE` can be done in constructor. When `hasHighDensity` changes, update rowSize in ngOnChanges / input setter. Consider only first change if that's what you want

eminecula07: I don't see the benefits here ... why not use ngOnInit but check at every change of hasHighDensity whether it is the first one or not?

CatalinB7: `UiGridComponent` did not implement `OnInit` before, that seems like a breaking change to me

eminecula07: From my experience with other frameworks/libs (Vue/React) I'd say it can't be called a breaking change, although it adds a slight complexity to the component.
I still believe that if we're interested only in the initial value of an input, we should use ngOnInit and not ngOnChanges even if it's the only reason for implementing it.
Maybe the others have other thoughts on this!?

alexdudescu: On the topic of breaking changes: I fail to see how this is a breaking change. If so please let's discuss @CatalinB7.

alexdudescu: Code wise: why are we doing this in ngOnInit and not in constructor?

projects/angular/components/ui-grid/src/ui-grid.component.ts
```diff
@@ -346,7 +356,7 @@ export class UiGridComponent<T extends IGridDataEntry>
      *
      */
     @Input()
-    rowSize: number;
+    rowSize!: number;
```
alexdudescu, line 359: You are marking this field, which is potentially undefined, as always having a value. It's not assigned in constructor either so how are we making sure this is defined?

projects/angular/components/ui-grid/src/ui-grid.component.ts
eminecula07: In the constructor we don't know anything about the input value of hasHighDensity!

projects/angular/components/ui-grid/src/ui-grid.component.ts
eminecula07: After ngOnInit is called it is surely defined! Moreover, as this is an input variable, we should not use it in the constructor; in the worst case we could use it in ngOnChanges (which is first called before ngOnInit), but it's not the case; and if it was, we could move the code in ngOnInit and add an extra check for rowSize in ngOnChanges.
Not sure what to say about this ... if you think we might need to use rowSize in ngOnChanges, then maybe it's worth removing ngOnInit; otherwise, I think it's fine to have this non-null assertion operator and the initialization code for rowSize in ngOnInit.

projects/angular/components/ui-grid/src/ui-grid.component.ts
CatalinB7: > I fail to see how this is a breaking change.

imagine a consumer having some class SpecialGrid that extends `UiGridComponent`. Let's say that SpecialGrid implements `ngOnInit`. After we change `UiGridComponent` to have its own `ngOnInit`, the class `SpecialGrid` will not function as expected. IMO it's a breaking change; i agree it's not something probable to encounter. On the topic of realistic use cases, who is going to change `hasHighDensity` input so we could not have the logic of `_setInitialRowSize` inside the setter?

alexdudescu: The code of a library is to be used as documented in it's API in a sort of a black-box fashion. You guarantee that as long as you call function x on class y it returns z. If you change something and it doesn't return z anymore in the same conditions as it did before, than that's a breaking change. However If (as a consumer) you are going to extend functionality and rely on a class that's not explicitly meant to be extended and your code breaks than congrats you just shot yourself in the foot and that's not our responsibility to prevent 🦶🏻 🔫

CatalinB7: the same train of thoughts as in the last sentence can also be applied to reason about possible value changes of an input that is not supposed to be changed. Anyway, i'm not fervent about not implementing new ng hooks to classes, it only seemed strange how the grid component could grow so much without needing to do it.

projects/angular/components/ui-grid/src/ui-grid.component.scss
alexdudescu: To be addressed in a future PR here: https://uipath.atlassian.net/browse/OR-75789

projects/angular/components/ui-grid/src/ui-grid.component.scss
alexdudescu: As above, to be resolved by: https://uipath.atlassian.net/browse/OR-75789

projects/angular/components/ui-grid/src/ui-grid.component.ts
```diff
@@ -1021,6 +1032,10 @@ export class UiGridComponent<T extends IGridDataEntry>
         return of(true);
     };
 
+    ngOnInit(): void {
```
alexdudescu, line 1035: This is needed after all since the rowSize's default value depends on highDensity

alexdudescu (approved):

CatalinB7 (approved):

PR #465, opened 2024-02-23
fix(grid): fix the height of the page size select

eminecula07:
https://uipath.atlassian.net/browse/OR-76366


AlexAndriesUiPath (approved):

eminecula07:
Will create another PR directly in Apollo.

